PR #9, opened 2020-04-22
WIP: support doc generation

KimBurgess:
Fixes an issue where running `crystal docs` would result in a compiler error and block doc generation. Functionality for build uses should be the same, but added a nicer compile time error for the missing env var too.


KimBurgess:
Closing this for now. Will re-open a new PR after the test harness has been split from the drivers.

Issue #2, opened 2020-03-24
Separate compilation logic into a shard


caspiano:
The logic used in [drivers](https://github.com/PlaceOS/drivers/tree/master/src/drivers) could be separated into a shard.

This logic is integral to the application, and its versioning is currently tied to PlaceOS's public driver offering.

PR #10, opened 2020-04-28
refactor: externalise compiler

KimBurgess:
Shifts all compiler components to [PlaceOS/compiler](https://github.com/PlaceOS/compiler), which is a pruned fork of this repo.

See #2


KimBurgess:
Next steps from here will be to separate the test-harness and the drivers themselves, but suggest we take this one step at a time to minimise chaos.

PR #14, opened 2020-06-17
build: support crystal 0.35.0

caspiano:
- Bump crystal in Dockerfile to 0.35.0
- Update dependencies to `placeos` scoped versions

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -1,5 +1,6 @@
 name: drivers
-version: 1.2.0
+version: 2.0.0
+crystal: 0.35.0
 
 # compile target
 targets:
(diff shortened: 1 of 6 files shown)


shard.yml
```diff
@@ -9,13 +10,13 @@ targets:
...
 
-  driver:
+  placeos-driver:
     github: placeos/driver
-    version: ~> 2.0
+    version: ~> 3.0
 
-  compiler:
+  placeos-compiler:
     github: placeos/compiler
```
KimBurgess, line 20: Is there a min version that can be specified for compiler?

PR #18, opened 2020-06-19
fix: set local shards path

### LINKED ISSUE #12: Specs don't run unless shards are installed on host
Due to `lib/` being mapped in as a volume, specs will not run until `shards install` has been run on the host.

To simplify initial spinup it may be worth mapping these in to an alternative path and appending this to the `CRYSTAL_PATH` env var.

This may cause some issues with handling different shard versions / introduce fragility where specs run locally, but not on other environments. Needs some further thought & discussion before implementation.

caspiano:
Set `CRYSTAL_PATH` environment variable in `docker-compose.yml`

Fixes #12

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,8 @@ services:
     restart: always
     container_name: drivers
     hostname: drivers
+    environment:
+      - CRYSTAL_PATH=/lib/local-shards
     depends_on:
       - redis
     ports:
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -14,6 +14,8 @@ services:
     restart: always
     container_name: drivers
     hostname: drivers
+    environment:
+      - CRYSTAL_PATH=/lib/local-shards
```
KimBurgess, line 18:
```suggestion
      - CRYSTAL_PATH=lib:/lib/local-shards
```
[Default behaviour](https://github.com/crystal-lang/crystal/blob/c41640fc205840a544f23eda33a6fedb4a184088/src/compiler/crystal/crystal_path.cr#L23) is to preference default path over custom unless specified explicitly in the ordering. In this case it's preferable to let users update/add shards externally and to run tests without container rebuilds.

PR #19, opened 2020-06-22
Port Samsung display driver

pkheav:


drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+    data.scan(/.{2}/) do |md|
+      output << md[0].to_i(16)
+    end
+
+    String.build do |io|
+      output.each do |number|
+        io.write_byte number.to_u8
+      end
+    end
+  end
```
KimBurgess, line 153: This can be swapped for the [implementation available from Transcoder](https://github.com/PlaceOS/driver/blob/master/src/placeos-driver/utilities/transcoder.cr).

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+    do_send("time", time_request)
+
+    state = enable ? "01" : "00"
+    vol = volume.to_s(16).rjust(2, '0')
+    # on 03:45 am enabled  off 03:30 am enabled  on-everyday  ignore manual  off-everyday  ignore manual  volume 15  input HDMI   holiday apply
+    custom_mdc(
+      "A4",
+      "03-2D-01 #{state}  03-1E-01   #{state}      01          80               01           80          #{vol}        21          01"
+    )
+  end
```
KimBurgess, line 191: This looks to be very specific to certain deploy requirements. For simplicity, drop this from the initial implementation. If / when there's a need to use the devices internal timer (over a trigger), this can be implementation in a more general way that allows defining all device timer actions.

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+    Widi          = 0x61
+    Magic_info    = 0x20
+    Whiteboard    = 0x64
+  end
+
+  def switch_to(input : String, **options)
+    self[:input_stable] = false
+    self[:input_target] = input
+    do_send("input", INPUTS.parse(input).value, **options)
+  end
```
KimBurgess, line 217: As with the power control, there is also a standard interface now for switching and input selection.

https://github.com/PlaceOS/driver/blob/master/src/placeos-driver/interface/switchable.cr

KimBurgess: Display will use the InputSelection interface within there.

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+    state = enable ? "01" : "00"
+    vol = volume.to_s(16).rjust(2, '0')
+    # on 03:45 am enabled  off 03:30 am enabled  on-everyday  ignore manual  off-everyday  ignore manual  volume 15  input HDMI   holiday apply
+    custom_mdc(
+      "A4",
+      "03-2D-01 #{state}  03-1E-01   #{state}      01          80               01           80          #{vol}        21          01"
+    )
+  end
+
+  enum INPUTS
```
KimBurgess, line 193: [Convention](https://crystal-lang.org/reference/conventions/coding_style.html) is to used camelcase for all type names.

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+      layout,             # layout mode (1..6)
+      SCALEMODE.parse(scale).value,  # scaling for main source
+      inputs.flat_map do |input|
+        [INPUTS.parse(input).value, SCALEMODE.parse(scale).value]
+      end
+    ].flatten
+
+    switch_to(main_source, **options)
+    do_send("screen_split", data, **options)
+  end
```
KimBurgess, line 240: This can be dropped from initial port, then added if/when needed.

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+
+  def in_range(val : Int32, max : Int32) : Int32
+    min = 0
+    if val < min
+      val = min
+    elsif val > max
+      val = max
+    end
+    val
+  end
```
KimBurgess, line 250: This can be replaced by [`#clamp` from stdlib](https://crystal-lang.org/api/0.35.1/Comparable.html#clamp(min,max)-instance-method)

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+      unmute_audio
+    end
+  end
+
+  def unmute_audio
+    if self[:audio_mute]?
+      self[:audio_mute] = false
+      volume(self[:previous_volume].as_i? || 50)
+    end
+  end
```
KimBurgess, line 270: More [interfaces](https://github.com/PlaceOS/driver/blob/master/src/placeos-driver/interface/muteable.cr) available for this too.

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+    "blue_gain"
+  ]
+
+  # Macro to define methods from the array above
+  {% for name in METHODS %}
+    def {{name.id}}(val : Int32, **options)
+      val = in_range(val, 100)
+      do_send({{name.id.stringify}}, val, **options)
+    end
+  {% end %}
```
KimBurgess, line 323: When using macros like this, unless it's needed elsewhere, you can just keep the collection that you're iterating over directly in the macro to keep the resulting code clean.

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+    if mac
+      # config is the database model representing this device
+      wake_device(mac, broadcast)
+      info = "Wake on Lan for MAC #{mac}"
+      info += " directed to VLAN #{broadcast}" if broadcast
+      logger.debug { info }
+    else
+      logger.debug { "No MAC address provided" }
+    end
+  end
```
KimBurgess, line 361: Wake on LAN is now provided as part of the [base driver](https://github.com/PlaceOS/driver/blob/60c4604e987fe541cfdbf0367a18f6da385890d2/src/placeos-driver.cr#L145-L147). This can be dropped.

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+    # Calculate checksum of response
+    checksum = data[1..-2].sum & 0xFF
+
+    # Pop also removes the checksum from the response here
+    if data.pop != checksum
+      logger.error { "Invalid checksum\nChecksum should be: #{checksum.to_s(16)}" }
+      return task.try &.retry
+    end
+
+    status = RESPONSESTATUS.new(data[4])
```
KimBurgess, line 382: In cases like this where you are wanting to strictly parse into defined Enum values, check out the `.from_value` and `from_value?` methods: https://play.crystal-lang.org/#/r/9b9v.

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+    command = COMMAND.new(data[5])
+    values = data[6..-1]
+    value = values.first
+
+    case status
+    when RESPONSESTATUS::Ack
```
KimBurgess, line 388: Really neat thing you get from enums: nice methods for query if a value is a specific type. In this case you can:
```crystal
case status
when .ack?
  # ...
```

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+    Net_standby      = 0xB5 # Keep NIC active in standby
+    Eco_solution     = 0xE6 # Eco options (auto power off)
+    Auto_power       = 0x33
+    Screen_split     = 0xB2 # Tri / quad split (larger panels only)
+    Software_version = 0x0E
+    Serial_number    = 0x0B
+    Time             = 0xA7
+  end
+
+  private def do_send(command : String | Int, data : Int | Array(Int) = [] of UInt8, **options)
```
KimBurgess, line 477: Use the types here to help prevent error. `command` should only accept an `Command` enum.

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+
+    # # options[:name] = command if data.length > 0 # name unless status request
+    command = COMMAND.parse(command).value if command.is_a?(String)
+
+    data = [command, @id, data.size] + data # Build request
+    data << (data.sum & 0xFF)               # Add checksum
+    data = [0xAA] + data                    # Add header
+
+    # Convert to Bytes
+    data = Slice.new(data.size) { |i| data[i].to_u8 }
```
KimBurgess, line 488: There's a quite a few intermediate arrays being created here. Take a looks at [this method for creating Slices](https://crystal-lang.org/api/0.35.1/Slice.html#new(size:Int,*,read_only=false,&)-class-method).

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+
+    logger.debug { "Sending to Samsung: #{data}}" }
+    logger.debug { "hexbytes: #{data.hexstring}" }
```
KimBurgess, line 491: Recommend dropping at least one of these.

drivers/samsung/displays/md_series.cr
pkheav: Thanks @KimBurgess , missed this one in the docs even though it was closer to the top of the page lol

drivers/samsung/displays/md_series.cr
pkheav: nice, thought there would be a better way to do this

drivers/samsung/displays/md_series.cr
pkheav: Hey @KimBurgess , I've made this change but there was a method in the ruby driver which allowed sending of custom commands via backoffice. Let me know if I should remove this though if it's not needed.

https://github.com/PlaceOS/drivers/blob/78c91c24b876507d49999657f36b0580ffa79214/drivers/samsung/displays/mdc_protocol.cr#L149-L154

drivers/samsung/displays/md_series.cr
pkheav: Hey @KimBurgess , I've made this change as I think it may be cleaner than creating a new slice and using a block to do things. Let me know what you think.

I haven't found a way to copy an array or even another Slice into a Slice so I've settled on doing this https://github.com/PlaceOS/drivers/commit/f5fc9e9798f632b4d832686969a3d4a8bddd0d0d#diff-96bf90ab842197ca263274d3fdb8e648R473

drivers/samsung/displays/md_series.cr
KimBurgess: Yep, nuke it. Better to have fully defined commands when / where they're needed.

drivers/samsung/displays/md_series.cr
KimBurgess: Nice! I really like that. Makes it super clear with the mapping to those different byte offsets in the protocol.

If you wanted to wrap this up a little further, you can shift the command assembly into the Command Enum itself and do away with do_send altogether. Something like:
```crystal
enum Command
  Power = 0x11
  Foo   = 0x01
  Bar   = 0x02
  Baz   = 0x03
  
  def build(*data)
    build id: 0xFE_u8, data: data
  end
  
  def build(id : UInt8, data : Tuple) : Bytes
    Bytes.new(data.size + 5).tap do |bytes|
      bytes[0] = 0xAA_u8
      bytes[1] = self.to_u8
      bytes[2] = id
      bytes[3] = data.size.to_u8
      data.each_with_index(4) { |b, i| bytes[i] = b.to_u8 }
      bytes[-1] = bytes[1..-2].reduce &.+.to_u8!
    end
  end
end
```
https://play.crystal-lang.org/#/r/9bp6

drivers/samsung/displays/md_series.cr
pkheav: No worries, nuked!
https://github.com/PlaceOS/drivers/pull/19/commits/52f408d53c0eedfd0201fd78dcdf27a7ec7ff8a6

drivers/samsung/displays/md_series.cr
pkheav: oh wow, that's neat. Thanks @KimBurgess and also for picking up that I wasn't calculating the checksum correctly. Didn't know how to use the reduce method in crystal so I just used the sum but and took the last 2 bytes which I can see now is not equivalent.

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+      end
+    end
+  end
+
+  # ability to send custom mdc commands via backoffice
+  def custom_mdc(command : String, value : String)
+    command = byte_to_hex(command).bytes[0].to_i
+    data = byte_to_hex(value).bytes
+    do_send(command, data)
+  end
```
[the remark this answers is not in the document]
pkheav: I've removed this method but good to know in future

drivers/samsung/displays/md_series.cr
pkheav: Hey @KimBurgess , I had to make this change on my local to get this working but I've also opened a PR for it.

https://github.com/PlaceOS/driver/pull/32/files

drivers/samsung/displays/md_series.cr
pkheav: removed!

drivers/samsung/displays/md_series.cr
pkheav: Hey @KimBurgess , I'm not sure how to make this any cleaner. I'm using `METHODS` in the first line of my macro so not sure where I would define the collection elsewhere.

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+      switch_to(@blank.as(String)) if @blank && self[:power]?
+      do_send("panel_mute", 1)
+    elsif !@rs232 && !self[:connected]?
+       wake(broadcast)
+    else
+      # Power on
+      do_send("hard_off", 1)
+      do_send("panel_mute", 0)
+    end
+  end
```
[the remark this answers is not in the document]
pkheav: Hey @KimBurgess , I'm not really sure how to use these both of the Powerable interface as well as this one https://github.com/PlaceOS/driver/blob/master/src/placeos-driver/interface/switchable.cr

drivers/samsung/displays/md_series.cr
pkheav: I've removed this as it's not actually needed

drivers/samsung/displays/md_series.cr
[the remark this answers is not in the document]
pkheav: addressed here https://github.com/PlaceOS/drivers/pull/19/commits/f5d723be148a62213aa94b6379187499d97326a5

drivers/samsung/displays/md_series.cr
pkheav: made this commit https://github.com/PlaceOS/drivers/pull/19/commits/cfd986bc02778c27a3cb4a76c3b9d86e8ee6f4f9

drivers/samsung/displays/md_series.cr
pkheav: https://github.com/PlaceOS/drivers/pull/19/commits/020b022dc19423575d1c2b045c61bc52b6eb9306

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,406 @@
...
+    "colour",
+    "tint",
+    "red_gain",
+    "green_gain",
+    "blue_gain"
+  ]
+
+  # Macro to define methods from the array above
+  {% for name in METHODS %}
+    def {{name.id}}(val : Int32, **options)
```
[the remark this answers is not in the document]
pkheav: https://github.com/PlaceOS/drivers/pull/19/commits/3d13aa1ba2579dfb6d6a7f5994838c5634260d30

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,406 @@
...
+    rs232_control: false
+  })
+
+  @id : Int32 = 0
+  @rs232 : Bool = false
+  @blank : String?
+
+  # TODO: figure out how to define indicator \xAA
+  def init_tokenizer
+    @buffer = Tokenizer.new do |io|
```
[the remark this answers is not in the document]
pkheav: https://github.com/PlaceOS/drivers/pull/19/commits/235de3f91da1a016bf82f1308702425a11c67b14

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,406 @@
...
+    end
+  end
+
+  def on_load
+    transport.tokenizer = init_tokenizer
+    on_update
+
+    self[:volume_min] = 0
+    self[:volume_max] = 100
+    self[:previous_volume] = 50
```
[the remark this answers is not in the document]
pkheav: ah okay yeah that makes sense. I've removed volume_min and volume_max since they're not used anywhere as well https://github.com/PlaceOS/drivers/pull/19/commits/5dad71750b639eb3ef8cd05ebb7d260d9c4dd0a4

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,406 @@
...
+
+    # Meta data for inquiring interfaces
+    self[:type] = :lcd
+    self[:input_stable] = true
+    self[:input_target] ||= :hdmi
+    self[:power_stable] = true
```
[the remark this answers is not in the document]
pkheav: https://github.com/PlaceOS/drivers/pull/19/commits/5dad71750b639eb3ef8cd05ebb7d260d9c4dd0a4

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,406 @@
...
+  end
+
+  def on_update
+    @id = setting(Int32, :display_id)
+    @rs232 = setting(Bool, :rs232_control)
+    @blank = setting?(String, :blanking_input)
+  end
+
+  def connected
+    do_poll
```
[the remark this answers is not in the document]
pkheav: Thanks, that's neat https://github.com/PlaceOS/drivers/pull/19/commits/aec8b98ccda6c35172f03e142c93c66a7cc63eb4

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,406 @@
...
+  # so if a nominal blank input is
+  # TODO: check type for broadcast is correct
+  def power(power : Bool, broadcast : String? = nil)
+    self[:power_target] = power
+    self[:power_stable] = false
+
+    if !power
+      # Blank the screen before turning off panel if required
+      # required by some video walls where screens are chained
+      switch_to(Inputs.parse(@blank.as(String))) if @blank && self[:power]?
```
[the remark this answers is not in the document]
pkheav: Thanks @KimBurgess , that's another neat tip https://github.com/PlaceOS/drivers/pull/19/commits/679e735edb8577e7af0fd0a9e2d26ff5456889c0

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,406 @@
...
+      do_send(Command::Panel_Mute, 1)
+    elsif !@rs232 && !self[:connected]? && broadcast
+       wake_device(broadcast.as(String))
```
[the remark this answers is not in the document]
pkheav: https://github.com/PlaceOS/drivers/pull/19/commits/532fd9b696fd4ab417151c2d17dad6afd523303f

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,406 @@
...
+  end
+
+  # Device auto power control (presumably signal based?)
+  def auto_power(enable : Bool, **options)
+    state = enable ? 1 : 0
+    do_send(Command::Auto_Power, state, **options)
+  end
+
+  # Display control
+  METHODS = [
```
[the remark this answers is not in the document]
pkheav: I've made this commit but not sure if there's a better/cleaner way of doing this https://github.com/PlaceOS/drivers/pull/19/commits/45c9e06bcbb70776e0383cbd643017b0cbbe580d

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,406 @@
...
+  default_settings({
+    display_id: 0,
+    rs232_control: false
+  })
+
+  @id : Int32 = 0
+  @rs232 : Bool = false
+  @blank : String?
+
+  # TODO: figure out how to define indicator \xAA
```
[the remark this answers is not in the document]
caspiano: [If the first byte is not `0xAA`, you can return a value `< 0` to flag that the message is malformed/incomplete](https://github.com/spider-gazelle/tokenizer#abstract)

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,389 @@
...
+    :green_gain,
+    :blue_gain,
+]
+
+  def do_device_config
+    logger.debug { "Syncronising device state with settings" }
+    DEVICE_SETTINGS.each do |name|
+      value = setting?(Int32, name)
+      # TODO: find out if these are equivalent
+      # __send__(name, value) unless value.nil?
```
caspiano, line 247:
```suggestion
```
I believe these are equivalent.

Pinging @stakach for confirmation.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,389 @@
...
+  @power_stable : Bool = true
+
+  def on_load
+    # TODO: figure out how to define indicator \xAA
+    transport.tokenizer = Tokenizer.new do |io|
+      bytes = io.peek
+      logger.debug { "Received: #{bytes}" }
+      # [header, command, id, data.size, [data], checksum]
+      bytes[3].to_i + 5
+    end
```
caspiano, line 69:
```suggestion
   
   MESSAGE_INDICATOR = 0xAA_u8
 
   def on_load
     transport.tokenizer = Tokenizer.new do |io|
       bytes = io.peek
       logger.debug { "Received: #{bytes}" }
       
       # Ensure header is well-formed
       return -1 unless bytes.first == MESSAGE_INDICATOR
       
       # [header, command, id, data.size, [data], checksum]
       bytes[3].to_i + 5
```

caspiano: This is incorrect, please disregard

drivers/samsung/displays/mdc_protocol.cr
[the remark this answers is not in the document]
pkheav: Thanks @caspiano , that is surprisingly simple and makes sense. https://github.com/PlaceOS/drivers/pull/19/commits/f04f5cb5f48e4adda9ab9b6235f5d8b0b887d54e

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -63,9 +63,12 @@ class Samsung::Displays::MDCProtocol < PlaceOS::Driver
     # TODO: figure out how to define indicator \xAA
     transport.tokenizer = Tokenizer.new do |io|
       bytes = io.peek
+      # disconnect if the first byte is not 0xAA
+      disconnect if bytes[0] != 170
```
caspiano:
```suggestion
    transport.tokenizer = Tokenizer.new do |io|
      bytes = io.peek
      # Ensure message indicator is well-formed
      disconnect unless bytes.first == 0xAA_u8
```

You could put 0xAA_u8 into a constant to reduce instances of magic values and it even clearer

pkheav: Thanks @caspiano , yeah that is much better https://github.com/PlaceOS/drivers/pull/19/commits/cf5698e201ffe2502c3a286d4b94f88433f2a277

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,393 @@
...
+    Auto_Power       = 0x33
+    Screen_Split     = 0xB2 # Tri / quad split (larger panels only)
+    Software_Version = 0x0E
+    Serial_Number    = 0x0B
+    Time             = 0xA7
+    Timer            = 0xA4
+
+    def build(id : Int32, data : Bytes) : Bytes
+      Bytes.new(data.size + 5).tap do |bytes|
+        bytes[0] = 0xAA_u8                              # Header
```
caspiano, line 372:
```suggestion
        bytes[0] = INDICATOR                            # Header
```

caspiano (dismissed):

drivers/samsung/displays/mdc_protocol.cr
KimBurgess: They’re not equivalent.

`__send__` is part of Ruby’s dynamic dispatch. It let’s you invoke an arbitrary method based on a name passed at runtime (https://ruby-doc.org/core-2.7.1/BasicObject.html#method-i-__send__). It’s also available as `send` (original implementation), but the double underscore alias was added for cases where an object has an actual object specific `send` method, which in the case of drivers in Engine, it does (https://docs.acaengine.com/developer-guide/drivers/device#transmitting-a-request).

So in the case above, the commented code would invoke the method called `name` and pass it value. As this is pulling from a constant, the equivalent in crystal in this context would be to use a macro:
```crystal
{% for name in DEVICE_SETTINGS %}
  %value = setting?(Int32, {{name.id}})
  {{name.id}}(%value) unless %value.nil?
{% end %}
```

Keep in mind though that crystal != ruby so you can likely come up with a neater implementation to satisfy the same requirements (query possibly defined device config from settings and push to device).

caspiano: ahh the ol dynamic method dispatch

pkheav: Hey @KimBurgess, thanks for that, good to know what that means in ruby!

I'm trying to write this in crystal and following your suggestion have made this commit to split the boolean and integer device settings but I can't seem to get this to work. I'm not sure if it's possible to use a macro within a method but any help would be great.

https://github.com/PlaceOS/drivers/pull/19/commits/f3969c2b978e08884fd9344ac8c3a57ced6674c8

pkheav: I've also tried using `value` instead of `%value` as I'm not sure what `%` actually means syntax wise which doesn't work either.

caspiano: `%value` creates a unique local macro variable. It essentially prefixes the variable by a random string so you don't end up with accidental variable shadowing within the same scope.

pkheav: fixed this, guess I just needed to take a 2nd look at this and stringify the setting names. Can you take another look at this @KimBurgess https://github.com/PlaceOS/drivers/pull/19/commits/5c868f64a3435147e7d698bdb932cd6fdd212072

KimBurgess: Cool. That’s coming together. With some of these changes it looks to have led to a split with the way you’re handling these settings methods and their application based on the type (`Bool` or `Int32`). Something that may be worth playing with is removing that split by changing the way you’re defining what settings are available.

```crystal
DEVICE_SETTINGS = {
  network_standby: Bool,
  volume: Int32,
}

{% for name, kind in DEVICE_SETTINGS %}
  def {{name.id}}(val : {{kind}})
    {% if kind.resolve == Bool %}
      val = val ? 1 : 0
    {% elsif kind.resolve == Int32 %}
      val = val.clamp 0, 100
    {% end %}
    do_send Command::{{name.id.upcase}}, val
  end
{% end %}

# etc - similar impl for applying these based on PlaceOS settings now that you have the type info available.
```

drivers/samsung/displays/mdc_protocol.cr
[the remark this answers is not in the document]
KimBurgess: LGTM.

If you’re worried about readability, you can still split that across multiple lines (similar to how it was defined in the constant).

drivers/samsung/displays/mdc_protocol.cr
KimBurgess: Hmmmm, that’s annoyingly unique. You could either keep that as a standalone method, or add in a special case within that macro. Neither is ideal, but I do think the concept of declaring all settings in one spot is worthwhile.

pkheav: yeah definitely agree on declaring them all in one spot, I've made this commit for that https://github.com/PlaceOS/drivers/pull/19/commits/ad373785f29e399586a1d5613e9d387419fa3855

drivers/samsung/displays/md_series.cr
```diff
@@ -0,0 +1,500 @@
...
+    do_send("hard_off", 0)
+    do_poll
+  end
+
+  # TODO: figure out what block is for
+  # def power?(**options, &block)
+  def power?(**options)
+    # options[:emit] = block unless block.nil?
+    do_send("panel_mute", [] of UInt8, **options)
+  end
```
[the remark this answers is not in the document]
stakach: yeah, doesn't really matter what that function is named, however I would keep it `power?` and have the function return the result of the query.

i.e. you can do 
```
def power?
  do_send(Command::Panel_Mute, Bytes.empty, **options).get
  self[:power]
end
```

Which would return Bool

drivers/samsung/displays/mdc_protocol.cr
pkheav: Hey @KimBurgess , can you give this another review when you can please?

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,376 @@
...
+  default_settings({
+    display_id: 0,
+    rs232_control: false
+  })
+
+  @id : Int32 = 0
+  @rs232 : Bool = false
+  @blank : Input?
+  @previous_volume : Int32 = 50
+  # Meta data for inquiring interfaces
```
KimBurgess, line 59: This comment doesn't look to apply any more.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,376 @@
...
+  @input_stable : Bool = true
+  @input_target : Input = Input::Hdmi
```
KimBurgess, line 61: Defaulting to an input here will result in a connected display being forced to HDMI after the first poll, despite this not being requested by the driver. May be worth making this a nilable field.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,376 @@
...
+
+  def disconnected
+    self[:power] = false unless @rs232
+    schedule.clear
+  end
+
+  # As true power off disconnects the server we only want to
+  # power off the panel. This doesn't work in video walls
+  # so if a nominal blank input is
+  # TODO: check type for broadcast is correct
```
KimBurgess, line 102: This TODO should be resolved before merge.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,376 @@
...
+    self[:power]
+  end
+
+  # Mutes audio + video
+  def mute(
+    state : Bool = true,
+    index : Int32 | String = 0,
+    layer : MuteLayer = MuteLayer::AudioVideo
+  )
+    logger.debug { "requested mute state: #{state}" }
```
KimBurgess, line 139: For consistency these debug messages should exist across all device interactions, or none.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,376 @@
...
+  def power?(**options)
+    do_send(Command::Panel_Mute, Bytes.empty, **options).get
+    self[:power]
+  end
+
+  # Mutes audio + video
+  def mute(
+    state : Bool = true,
+    index : Int32 | String = 0,
+    layer : MuteLayer = MuteLayer::AudioVideo
```
KimBurgess, line 137: `layer` should be used to control if this targets audio, video or both.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,376 @@
...
+
+  def power?(**options)
+    do_send(Command::Panel_Mute, Bytes.empty, **options).get
+    self[:power]
+  end
+
+  # Mutes audio + video
+  def mute(
+    state : Bool = true,
+    index : Int32 | String = 0,
```
KimBurgess, line 136: `layer` should be used and an `ArgumentError` raised if un unsupported index is passed.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,376 @@
...
+    state = state ? 1 : 0
+    do_send(Command::Panel_Mute, state)
+  end
+
+  def unmute_video
+    mute_video(false)
+  end
+
+  # Emulate mute
+  def mute_audio(val : Bool = true)
```
KimBurgess, line 155: Suggest renaming `val` to `state`.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,376 @@
...
+
+  def unmute_audio
+    if self[:audio_mute]?
+      self[:audio_mute] = false
+      volume(@previous_volume)
+    end
+  end
+
+  # check software version
+  def software_version?
```
KimBurgess, line 175:
```suggestion
  def software_version
```

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,376 @@
...
+    do_send(Command::Software_Version)
+  end
+
+  def serial_number?
```
KimBurgess, line 179:
```suggestion
  def serial_number
```

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,376 @@
...
+    @input_target = input
+    do_send(Command::Input, input.value, **options)
+  end
+
+  enum SpeakerModes
+    Internal = 0
+    External = 1
+  end
+
+  def speaker_select(mode : String, **options)
```
KimBurgess, line 194: This should likely take a `SpeakerModes` as the argument type.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,376 @@
...
+
+    case status
+    when .ack?
+      case command
+      when .status?
+        self[:hard_off]   = values[0] == 0
+        self[:power]      = false if self[:hard_off]
+        self[:volume]     = values[1]
+        self[:audio_mute] = values[2] == 1
+        self[:input]      = Inputs.from_value(values[3]).to_s
```
KimBurgess, line 271: I _think_ that enum values should have `#to_s` invoked automatically. If this is the case, you should be safe to drop the `.to_s` here along with other spots that state is being published.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,376 @@
...
+      end
+    end
+  end
+
+  private def do_send(command : Command, data : Int | Bytes = Bytes.empty, **options)
+    data = Bytes[data] if data.is_a?(Int)
+    bytes = command.build(@id, data)
+    logger.debug { "Sending to Samsung: #{bytes.hexstring}" }
+    send(bytes, **options)
+    # TODO: find out if this is necessary
```
KimBurgess, line 370: TODO needs to be resolved before merge.

drivers/samsung/displays/mdc_protocol_spec.cr
```diff
@@ -0,0 +1,34 @@
...
+  # power? will take priority over status as status has priority = 0
+  # power? -> panel_mute
+  should_send("\xAA\xF9#{id}\x00\xF9")
+  responds("\xAA\xFF#{id}\x03A\xF9\x00\xFF")
+  status[:power].should eq(true)
+  # status
+  should_send("\xAA\x00#{id}\x00\x00")
+  responds("\xAA\xFF#{id}\x09A\x00\x01\x06\x00\x14\x00\x00\x00\xFF")
+  status[:hard_off].should eq(false)
+  # status[:power].should eq(true)
```
KimBurgess, line 16: Is there a reason this is commented out?

drivers/samsung/displays/mdc_protocol_spec.cr
```diff
@@ -0,0 +1,34 @@
...
+
+  exec(:volume, 24)
+  should_send("\xAA\x12#{id}\x01\x18\x12")
+  responds("\xAA\xFF\x00\x03A\x12\x18\xFF")
+  status[:volume].should eq(24)
+  status[:audio_mute].should eq(false)
+
+  exec(:mute, true)
+  responds("\xAA\xFF#{id}\x03A\xF9\x01\xFF")
+  status[:power].should eq(false)
```
KimBurgess, line 29: Why is power state being tested with the `mute` spec?

drivers/samsung/displays/mdc_protocol_spec.cr
```diff
@@ -0,0 +1,34 @@
...
+
+  exec(:unmute)
+  responds("\xAA\xFF#{id}\x03A\xF9\x00\xFF")
+  status[:power].should eq(true)
```
KimBurgess, line 33: As above. Doesn't look like this is the correct status to be checking.

drivers/samsung/displays/mdc_protocol_spec.cr
```diff
@@ -0,0 +1,34 @@
...
+end
```
KimBurgess, line 34: Would be good to add some more specs here to ensure all core device functionality is tested: power, voluming, muting, input selection etc. While these _may_ work today, having this covered allows for safer refactoring and additions by future contributors.

KimBurgess (changes requested):

drivers/samsung/displays/mdc_protocol.cr
pkheav: I've removed broadcast as that was from the old ruby implementation. Can add it back later if it's needed. https://github.com/PlaceOS/drivers/pull/19/commits/630d1213b538f2702af767969d2bd9a33d6597d7

drivers/samsung/displays/mdc_protocol_spec.cr
pkheav: This is because I think the spec runner doesn't preserve the previous device state set here https://github.com/PlaceOS/drivers/blob/59fe9aa630abac79aa710ade043602e70706652c/drivers/samsung/displays/mdc_protocol_spec.cr#L11

pkheav: actually it's I think because this line get's run if the setting exists even though `self[:hard_off] == false` https://github.com/PlaceOS/drivers/pull/19/commits/3781786578535012b59047db762ceec920412bb5#diff-96bf90ab842197ca263274d3fdb8e648L263

pkheav: made another commit which is probably more ideal and works with crystal better https://github.com/PlaceOS/drivers/pull/19/commits/95f64ba271b69c74ebc321d2e158dda0b42fcaf7

drivers/samsung/displays/mdc_protocol_spec.cr
pkheav: I think the main reasoning behind this is because `self[:power]` is representative of whether video is actually being displayed on the tv/projector. Pretty sure this was an intentional design choice @stakach made according to these comments.

https://github.com/PlaceOS/drivers/blob/95f64ba271b69c74ebc321d2e158dda0b42fcaf7/drivers/samsung/displays/mdc_protocol.cr#L98
https://github.com/PlaceOS/drivers/blob/95f64ba271b69c74ebc321d2e158dda0b42fcaf7/drivers/samsung/displays/mdc_protocol.cr#L104-L105

pkheav: which is why `def power` uses `Command::Panel_Mute`

drivers/samsung/displays/mdc_protocol.cr
pkheav: removed as it's from the old ruby implementation

drivers/samsung/displays/mdc_protocol.cr
pkheav: I'm not sure what you mean by this one @KimBurgess

pkheav: never mind, think I worked out how to use the Muteable interface properly

https://github.com/PlaceOS/drivers/pull/19/commits/14468b4df0f578f7efe54ff4d0edd0ff45cf95c7

drivers/samsung/displays/mdc_protocol_spec.cr
pkheav: I've added more specs but I'm not sure on how to actually specify an Input to add a test for that in the spec.

pkheav:
Hopefully everything is ready to go now @KimBurgess

drivers/samsung/displays/mdc_protocol_spec.cr
pkheav: bump on this @KimBurgess , if you can take a look when possible please

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,364 @@
...
+    self[:power]
+  end
+
+  # Mutes both audio/video
+  def mute(
+    state : Bool = true,
+    index : Int32 | String = 0,
+    layer : MuteLayer = MuteLayer::AudioVideo
+  )
+    mute_video(state) unless layer == MuteLayer::Audio
```
KimBurgess, line 133: Rather than using an negative condition, this is better suited to matching a specific layer (either an `if` or a `case` statement). This way if other items are added to the MuteLayer enum, it won't impact behaviour here.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,364 @@
...
+        self[:serial_number] = values.join
+      else
+        logger.debug { "Samsung responded with ACK: #{value}" }
+      end
+
+      task.try &.success
+    when .nak?
+      task.try &.abort("Samsung responded with NAK: #{hex}")
+    else
+      task.try &.abort("Samsung aborted with: #{hex}")
```
KimBurgess, line 301: If the response is neither a nak or ack, it may be worth retrying to task instead of aborting.

drivers/samsung/displays/mdc_protocol_spec.cr
```diff
@@ -0,0 +1,55 @@
+  # [header, command, id, data.size, [data], checksum]
+require "placeos-driver/interface/muteable"
```
KimBurgess, line 2: This require shouldn't be needed here.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,364 @@
...
+  def on_update
+    @id = setting(Int32, :display_id)
+    @rs232 = setting(Bool, :rs232_control)
+    if blanking_input = setting?(String, :blanking_input)
+      @blank = Input.parse?(blanking_input)
+    end
+  end
+
+  def connected
+    do_device_config unless (hard_off = self[:hard_off]?) && hard_off.as_bool?
```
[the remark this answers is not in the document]
pkheav: This doesn't work if the self[:whatever_key] = false though which I discovered when the driver wasn't behaving as expected when running the spec. I've added this at the bottom of my driver to test it out as well.

image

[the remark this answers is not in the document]
pkheav:
```  def test_self_truthy
    self[:test] = false
    logger.debug { "self[:test] = #{self[:test]}" }
    logger.debug { "type of self[:test] = #{self[:test]}" }


    if self[:test]?
      logger.debug { "self[:test]? is truthy" }
    else
      logger.debug { "self[:test]? is not truthy" }
    end

    if self[:test]?.try &.as_bool?
      logger.debug { "self[:test]?.try &.as_bool? is truthy" }
    else
      logger.debug { "self[:test]?.try &.as_bool? is not truthy" }
    end
  end```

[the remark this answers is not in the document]
Comment (pkheav) [reply drivers/samsung/displays/mdc_protocol.cr:86]: `(hard_off = self[:hard_off]?) && hard_off.as_bool?` this is equivalent to `self[:hard_off]?.try &.as_bool?` now that I know what `try` does in crystal

File: drivers/samsung/displays/mdc_protocol_spec.cr (hunk 49 above)
Comment (pkheav) [reply drivers/samsung/displays/mdc_protocol_spec.cr]: Sorry that was remnants of trying to test the mute function by passing MuteLayer in the spec

**Comment by pkheav** [comment]
Hey @KimBurgess , just re requested your review. I've addressed all your changes except for this one https://github.com/PlaceOS/drivers/pull/19#discussion_r496634729

File: drivers/samsung/displays/mdc_protocol.cr (hunk 41 above)
Comment (pkheav) [reply drivers/samsung/displays/mdc_protocol.cr:255]: oh yes that appears to be the case. Thanks, that'll be a nice cleanup

File: drivers/samsung/displays/mdc_protocol.cr [hunk 51]
```diff
@@ -0,0 +1,354 @@
...
+  end
+
+  def serial_number
+    do_send(Command::SerialNumber)
+  end
+
+  def switch_to(input : Input, **options)
+    @input_stable = false
+    input_target = @input_target
+    @input_target = input_target if input_target
```
KimBurgess, line 165: Is this the correct value being assigned to `@input_target`?

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,354 @@
...
+
+  enum ResponseStatus
+    Ack = 0x41 # A
+    Nak = 0x4e # N
+  end
+
+  def received(data, task)
+    hex = data.hexstring
+    logger.debug { "Samsung sent: #{hex}" }
+    data = data.map(&.to_i).to_a
```
KimBurgess, line 225: Is there a reason to remap each of these bytes to an int32?

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,354 @@
...
+
+    # Calculate checksum of response
+    checksum = data[1..-2].reduce(&.+)
+
+    # Pop also removes the checksum from the response here
+    if data.pop != checksum
+      logger.error { "Invalid checksum\nChecksum should be: #{checksum.to_s(16)}" }
```
KimBurgess, line 232: Try to avoid line breaks in error messages if possible.

drivers/samsung/displays/mdc_protocol.cr
```diff
@@ -0,0 +1,354 @@
...
+
+      task.try &.success
+    when .nak?
+      task.try &.abort("Samsung responded with NAK: #{hex}")
+    else
+      task.try &.retry
+    end
+  end
+
+  def check_power_state
```
KimBurgess, line 295: This looks to be more of an internal method. Can it be flagged as private?

KimBurgess (commented):
Only a couple of super small point for discussion here. Let's sort them, check all the spec pass then gtg!

drivers/samsung/displays/mdc_protocol.cr
pkheav: I think I was calculating the checksum incorrectly before and the result of that couldn't be stored in a UInt8 but now that you helped me sort that out, I don't actually need to do this. Thanks for picking this up, I've made this commit.
https://github.com/PlaceOS/drivers/pull/19/commits/886d433be586e9c4c28d9d049937dade3980107c

KimBurgess (commented):
Looking good - confirming specs are green here.

Before me merge as one last thing can you add some specs that cover `power` and `switch_to`.

KimBurgess (approved):
Nice one. Things should work in this state.

Before merge I'm opening up a discussion here with @stakach and @caspiano. There's some behaviour in this driver that's a carry over from the old one: both the power state and input are forced to specific targets based on the last applied state. If these change from an external source, the target states will be reapplied when this is detected.

IMHO this is something that we should make a rule to disallow (or at least strongly discourage) at the device driver level. Doing so will keep the drivers simpler so they purely provide control and an accurate view of the current state. These other behaviours can then be abstracted into logic drivers and applied where needed. It also removes some of potential failure modes of having duplicate module instances, other systems, or humans interacting with a physical device competing against each other. Thoughts?

KimBurgess:
@stakach pump on the query above.

stakach:
The way it works in the ruby version, I believe, is that the state is forced if selected by the driver.
Once in a stable state then it no longer forces the input and external changes are OK

The reason for doing this is that the display ignores commands, or is busy when powering on etc
and should go to the requested state as soon as it can

PR #24, opened 2020-07-14
Lenel OpenAccess driver

kimburgess:
Adds support for integration with [Lenel OnGuard](https://www.lenel.com/products/onguard) physical security / access control systems via the [Open Access](https://www.lenel.com/solutions/open-integration/oaap) API.

_Note:_ integration is currently in progress. PR opened for early discussion during testing / staging deployment.


KimBurgess:
Running into some issue with this during testing with a live system.

Following debugging, the `Lnl_Vistor` model shown within the API docs does not reflect what's exposed via the API. Notably, the `email` attribute is missing. The presents as queries using an `email = ...` filter failing due to invalid filter.

Output from a query to the `/type?version=1.0&type_name=Lnl_Vistor` endpoint on the system being used for testing here: https://gist.github.com/KimBurgess/cd43ecbe2813fd26600ee1cf58b84784.

Have reached out to Lenel for technical assistance.

KimBurgess:
Still awaiting vendor resolution on this. They have confirmed the email attribute is not present on the system currently being tested against. From communication so far this appears to be incorrect behaviour so will await resolution there rather than attempting to key via an alternative visitor attribute / set of attributes.

w-le:
In my testing just now, looks like create_cardholder() returns the `id` key as all caps instead of lowercase
Example real exec result:
```
{
    "ID": 23524
}
```

while `lookup_cardholder()` returns the id as lowercase as expected:
```
{
    "id": 23523,
    "firstname": "William",
    "lastname": "Le",
    "email": "will@place.tech"
}
```
Can you make the client downcase all the keys in create responses?
https://github.com/PlaceOS/drivers/pull/24/files#diff-47674f68012f1934fd127acdd4fae7fda86068d484fab73fff821f134b69a11aR102

I would suggest the actual code change required but I actually don't know because I'm not good enough at crystal yet to fully understand the syntax being used in that func. My guess is there would be a basic transform on Line 102 above, something like `.transform_keys { |k| k.downcase }` ?

kimburgess:
Good catch! Currently the create methods return back the API response transparently. There are [inconsistencies in the OpenAccess API](https://github.com/PlaceOS/drivers/blob/a28a1e244ac05505b9992e271dbdc7d2bd8de905/drivers/lenel/open_access/models.cr#L23-L26) so definitely makes sense to shield downstream users from this. One moment - I'll pop together and change and drop it through.

kimburgess:
@w-le not too much of an issue here as this will all be squashed before merge, but would be good to avoid commits like 73f332d making there way to any public branches.

w-le (approved):
Tested working OK against a real system.
Very elegant driver, it's seriously inspiring and will be a reference for my own dev

PR #50, opened 2020-09-10
Nec port

pkheav:


pkheav:
Okay I’ve just about finished writing tests now for the nec display driver which passes on both the ruby driver and the crystal driver. I’ll open a PR so you can review this @KimBurgess  but this driver has been quite a headache to write just because of the way the protocol has been implemented and how difficult it is to write tests for the spec.

Do let me know if anything can be cleaned up or done better though.

https://github.com/acaprojects/ruby-engine-drivers/commit/ee1ff475aed1852cdabcac050d26dc434a6b2383

drivers/nec/display/all.cr
```diff
@@ -0,0 +1,410 @@
...
+    @target_input = input
+    @target_audio = nil
+
+    data = OPERATIONS[:video_input]
+    data += input.value.to_s(16).upcase.rjust(4, '0')
+
+    do_send("set parameter", data, name: "input", delay: 6.seconds)
+    video_input
+
+    # Double check the input again!
```
KimBurgess, line 145: Is there a need for this? Command will auto-retry if unsuccessful.

drivers/nec/display/all.cr
```diff
@@ -0,0 +1,410 @@
...
+
+  def connected
+    schedule.every(50.seconds, true) do
+      do_poll
+    end
+  end
+
+  def disconnected
+    # Disconnected may be called without calling connected
+    # Hence the check if timer is nil here
```
KimBurgess, line 103: Again, any comments like this that do no contain information important to the operation of this specific driver can be dropped.

drivers/nec/display/all.cr
```diff
@@ -0,0 +1,410 @@
...
+    schedule.clear
+  end
+
+  def power(state : Bool)
```
KimBurgess, line 107: It's as little unclear what's happening in here. This method to be changing behaviour based on the current power state vs an intended state.

drivers/nec/display/all.cr
```diff
@@ -0,0 +1,410 @@
...
+        # auto_setup
+        # nothing needed to do here (we are delaying the next command by 4 seconds)
+      else
+        logger.info { "-- NEC LCD, unknown response: #{data[10..13]}" }
+        logger.info { "-- NEC LCD, full response was: #{data}" }
+      end
+  end
+
+  # Types of messages sent to and from the LCD
+  MSG_TYPE = {
```
KimBurgess, line 346: This should be able to become an `Enum`.

drivers/nec/display/all.cr
```diff
@@ -0,0 +1,410 @@
...
+    "command" => 'A',
+    'B' => "command_reply",
+    "get_parameter" => 'C',
+    'D' => "get_parameter_reply",
+    "set_parameter" => 'E',
+    'F' => "set_parameter_reply"
+  }
+
+  OPERATIONS = {
```
KimBurgess, line 355: Ditto with this one (encoding the current string values as the integers they represent).
```crystal
enum Operation
  VideoInput = 0x0060
  AudioInput = 0x022e
  # ...etc
end
```

drivers/nec/display/all.cr
```diff
@@ -0,0 +1,410 @@
...
+    end
+
+    logger.debug { "NEC LCD responded with hex #{hex}" }
+
+    command = MSG_TYPE[hex[4]]
+
+    case command # Check the MSG_TYPE (B, D or F)
+      when "command_reply"
+        # Power on and off
+        if hex[10..15] == "C203D6" # Means power comamnd
```
KimBurgess, line 242: It would be good to encode all of these 'magic numbers' within an enum or similar.

This driver looks to have left a little to be desired from it's original implementation. I'd recommend using the protocol manual as the main reference, and using that to help build abstractions that work well here as opposed to focusing on a line-by-line port of the ruby driver.

KimBurgess (changes requested):
There's a few things to neaten up here. Most of this is likely due to some pretty poor structure in the old implementation vs anything new that you've introduced. A main part of this porting process is neatening up these drivers so that they're something that's maintainable as opposed to something that works.

Quick note on naming too: this can just be `drivers/nec/display.cr`.

pkheav:
Hey @KimBurgess , I've been working through your feedback but I feel that using enums to store values makes the code a lot less readable mainly because the commands can be built using ascii characters which is easier to read like in the original ruby implementation. Let me know if you still think differently but I've made these 2 commits which aren't as readable imo.

https://github.com/PlaceOS/drivers/pull/50/commits/4301a0109be5e31628a0b0ef03246eddb6def428
https://github.com/PlaceOS/drivers/pull/50/commits/e0363f99a60fa1fcf7a8fcfd01f2932cf2c03d86

pkheav:
Actually just realised I converted those ascii incorrectly to the equivalent hex numbers so it should look more like this.
https://github.com/PlaceOS/drivers/pull/50/commits/a2819a0c6ea9ba077d00038525d3081fdcbd3146#

KimBurgess:
What a horrendously inefficient protocol! It has all the overhead of something easily typed by a human with the string encoding, but none of the readability.

One thing that will really help out is explicitly specifying a base when you're converting between string and numeric representations. By default, when you use `#to_s` and `#to_i` this assumes base 10, but you can still use these to map between different forms too.
```crystal
0xBEEF # => 48879

48879.to_s(16) # => "beef"

"beef".to_i(16) # => 48879
```

This should let you model some of the message structure in enums still, while keeping some sanity.

drivers/nec/display/all.cr
pkheav: thanks @KimBurgess for the to_s(16)/to_i(16) idea, that does help clean up things a bit
https://github.com/PlaceOS/drivers/pull/50/commits/d0146a1fbb04bdc787182cff378aae0231a6624b

pkheav: also this commit which kind of neatens things maybe
https://github.com/PlaceOS/drivers/pull/50/commits/65744ac9f4abf981d6116f16a4904ee43f586866

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,397 @@
...
+  @target_audio : Audio? = nil
+  @input_double_check : PlaceOS::Driver::Proxy::Scheduler? = nil
+
+  DELIMITER = 0x0D_u8
+
+  def on_load
+    # Communication settings
+    queue.delay = 120.milliseconds
+    # 0x0D (<CR> carriage return \r)
+    transport.tokenizer = Tokenizer.new(Bytes[DELIMITER])
```
KimBurgess, line 60:
```suggestion
    transport.tokenizer = Tokenizer.new(Bytes[DELIMITER])
```

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,397 @@
...
+      do_send(MsgType::Command, data, name: "power", delay: 5.seconds)
+      self[:warming] = true
+      self[:power] = true
+      logger.debug { "-- NEC LCD, requested to power on" }
+
+      power_on_delay
+      mute_status(20)
+      volume_status(20)
+    end
+  end
```
KimBurgess, line 100: This method still needs some attention. The `state` param does not look to be used anywhere.

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,397 @@
...
+  def power(state : Bool)
+    data = "C203D6"
+    current = self[:power]?
+
+    if current
+      data += "0004" # 0004 = Power Off
+      do_send(MsgType::Command, data, name: "power", delay: 10.seconds, timeout: 10.seconds)
+
+      self[:power] = false
+      logger.debug { "-- NEC LCD, requested to power off" }
```
KimBurgess, line 88: When adding debug message like this, it's a good idea to place them before the event they're referring to. This makes debugging significantly easier as if there's an issue, the action that led to it will appear in logs before hand, rather than possibly being dropped due to an exception.

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,397 @@
...
+
+class Nec::Display::All < PlaceOS::Driver
+  include Interface::Powerable
+  include Interface::AudioMuteable
+
+  enum Inputs
+    Vga          = 1
+    Rgbhv        = 2
+    Dvi          = 3
+    Hdmi_set     = 4
```
KimBurgess, line 16: Either `CamelCase` or `SCREAMING_CASE` for enum entries. Doesn't matter which one, just as long as it's consistent. 

```suggestion
    HdmiSet      = 4
```

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,397 @@
...
+    # @input_double_check = schedule.in(4.seconds) do
+    #   @input_double_check = nil
+    #   video_input
+    # end
+    # TODO: check if the below is equivalent to what's above
+    @input_double_check.try &.terminate
+    schedule.in(4.seconds) do
+      @input_double_check = nil
+      video_input.as(PlaceOS::Driver::Transport)
+    end
```
KimBurgess, line 126: As per the previous review - this can be dropped. If there's issue with a device not correctly responding, this should be handles as part of the retry behaviour for the input query.

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,397 @@
...
+
+    command = MsgType.from_value(data[4])
+
+    case command # Check the MsgType (B, D or F)
+    when .command_reply?
+      # Power on and off
+      if ascii_string[10..15] == "C203D6" # Means power comamnd
+        # 8..9 == "00" means no error
+        if ascii_string[8..9] == "00"
+          power_on_delay(99) # wait until the screen has turned on before sending commands (99 == high priority)
```
KimBurgess, line 213: This method doesn't look to exist anywhere.

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,397 @@
...
+    logger.debug { "requested to update mute to #{state}" }
+  end
+
+  def unmute_audio
+    mute_audio(false)
+  end
+
+  # LCD Response code
+  def received(data, task)
+    ascii_string = String.new(data)
```
KimBurgess, line 197: You may find it useful to use [`String#hexbytes`](https://crystal-lang.org/api/0.35.1/String.html#hexbytes:Bytes-instance-method) here. This will allow you to map back from the hex encoded ascii string, to the byte array that it presents.

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,397 @@
...
+  end
+
+  def self.format_value(value : Int, length : Int = 4) : String
+    value.to_s(16).upcase.rjust(length, '0')
+  end
+
+  {% for name in Command.constants.map(&.downcase) %}
+  @[Security(Level::Administrator)]
+    def {{name.id}}(priority : Int32 = 0)
+      do_send(MsgType::Get_parameter, Command.parse({{name}}), priority: priority, name: {{name.id}})
```
KimBurgess, line 346: You shouldn't need any runtime parsing here. This can just be `Command::{{name.id}}` if you move the name mapping to the method naming only.

If you end up changing those enum names to camelcase too, I'd recommend [`String#underscore`](https://crystal-lang.org/api/0.35.1/String.html#underscore(options:Unicode::CaseOptions=:none):String-instance-method) for helping with this.

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,397 @@
...
+    Auto_setup        = 0x001E
+    Power_query       = 0x01D6
+
+    def to_s : String
+      Nec::Display::All.format_value(self.value)
+    end
+  end
+
+  def self.format_value(value : Int, length : Int = 4) : String
+    value.to_s(16).upcase.rjust(length, '0')
```
KimBurgess, line 340: [`String#to_s`](https://crystal-lang.org/api/0.35.1/Int.html#to_s(base:Int=10,*,upcase:Bool=false):String-instance-method) provides an `upcase` param to save a double string creation here.

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,397 @@
...
+      # Check the checksum equals the second last element
+      checksum == data[-2]
+    else
+      true
+    end
+  end
+
+  # Builds the command and creates the checksum
+  # data can be an ascii encoded string
+  private def do_send(type : MsgType, data : Command | String, **options)
```
KimBurgess, line 366: Rather than passing `data` in as a string, I'd recommend keeping everything as enum's, or byte arrays until they reach here. That way you can perform all the mapping needed to convert to the hex string in only place.

drivers/nec/display/display_spec.cr
```diff
@@ -0,0 +1,5 @@
+DriverSpecs.mock_driver "Nec::Display::All" do
+  should_send("\x010*0A06\x0201D6\x03\x1F\x0D")
+     responds("\x0100*B12\x020200D60000040001\x03\x1F\x0D")
+  status[:power].should eq(true)
+end
```
KimBurgess, line 5: Before merging, this will need some more specs.

KimBurgess (changes requested):
Please don't take these review as me being overly critical of you - a lot of the drivers you are porting are not a great base to be working from.

I'm intentionally setting some ~significantly higher~ actually present base quality thresholds for anything moving across.

drivers/nec/display/display.cr
pkheav: sorry did not notice this, good pick up

drivers/nec/display/display.cr
pkheav: yep looks a lot more pleasing in camel case

pkheav: Oh I think doing this https://github.com/PlaceOS/drivers/pull/50/commits/90fc56dbd60dfba97f0d033885fc9337d3391f36#diff-53e26f291e6267f24d41f88df5c2351fR323-R332

means that functions created from this macro will be like `def videoinput` instead of `def video_input` https://github.com/PlaceOS/drivers/pull/50/commits/90fc56dbd60dfba97f0d033885fc9337d3391f36#diff-53e26f291e6267f24d41f88df5c2351fR343-R348

Is it okay to use underscores in the enum entries for this case then @KimBurgess ?

pkheav: nevermind I should of read your comment below https://github.com/PlaceOS/drivers/pull/50#discussion_r489943454

drivers/nec/display/display.cr
pkheav: I don't think it's possible to make `data` only take in an enum especially for the input switching commands as they both need a command enum and the actual input enum like here

https://github.com/PlaceOS/drivers/blob/e7f3d475fd30365e5e91ac93b94ed0a1081cabf3/drivers/nec/display/display.cr#L110

which data would evaluate to `"00600001"` if `input == Input::Vga`. I can convert `data` to bytes before giving it to `do_send` but seems easier to allow data to also take a string. Let me know if I'm going about this the wrong way or if there's a better way of doing this.

drivers/nec/display/all.cr
pkheav: I've got no idea but it's from the old ruby code so thought it might be needed. I've removed it for now and we can add it back in later if there are issues. https://github.com/PlaceOS/drivers/pull/50/commits/bad36ef655678893d95b21b2325362326df8980c

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,397 @@
...
+    Dvd1         = 12
+    Option       = 13
+    Dvd2         = 14
+    Display_port = 15
+    Hdmi         = 17
+    Hdmi2        = 18
+    Hdmi3        = 130
+    Usb          = 135
+
+    def to_s : String
```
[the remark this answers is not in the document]
pkheav: Hey @KimBurgess , I'm not sure what you mean by this

drivers/nec/display/display.cr
pkheav: I'm not sure how I can use that. Isn't data of type `Bytes`? 
image

drivers/nec/display/display.cr
pkheav: Okay I've made this commit which might still not be the best way but I couldn't figure out a way to only call `underscore` once and I can't use do `name: {{name.id}}` without getting an error in the driver tester

https://github.com/PlaceOS/drivers/pull/50/commits/c78a02894f37b391da953e4799600c6159e808e5

drivers/nec/display/display.cr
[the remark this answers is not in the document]
KimBurgess: This was in reference to `Inputs#to_s`. That serialisation is specific to how an `Input` should appear as part of the outgoing command. To keep this all grouped together, this can be performed within your `do_send` method.

drivers/nec/display/display.cr
KimBurgess: With this protocol the incoming data will be a `Bytes` containing an ASCII representation of the actual response packet. If you first read this into a string, then dump back out, this should decode into a suitable format for parsing.

```crystal
response = String.new(data).hexbytes
```

There's a more efficient way to do this which drop the need for that intermediate string, but the readability tradeoff is likely not worth the minuscule perf boost.

drivers/nec/display/display.cr
KimBurgess: The `name` param is used by the underlying Queue to make sure only one instance of the associated command is stacked up at any point. It just needs to be something that can be mapped into a `String`.

At the moment when that macro expands you will be left with `name: video_input` etc. `video_input` is not a valid identifier within you current scope so the compiler will complain about that. Tack on a [`.symbolize`](https://crystal-lang.org/api/0.35.1/Crystal/Macros/ASTNode.html#symbolize:SymbolLiteral-instance-method) or [`.stringify`](https://crystal-lang.org/api/0.35.1/Crystal/Macros/ASTNode.html#stringify:StringLiteral-instance-method) and it will expand to what you're likely after there.

Also, don't stress a great deal about perf / minimising operations within macros. While this is definitely a good goal, it's a cost which is only incurred once as the driver compiles. It's only the expanded code that contributes to run-time cost.

drivers/nec/display/display.cr
KimBurgess: See [other comments](https://github.com/PlaceOS/drivers/pull/50#discussion_r492443003) about reducing the number different representation formats that you are dealing with at any time.

The transport for this protocol sucks, but you have two clean delineation points where you can limit the impact of this oddity:
1. `do_send`
2. `received`

If you keep all mapping to/from the ascii encoded hex string in these places, all other sections of your code can deal with the protocol in it's direct format, without worrying about conversion, or mixing converted and non converted components.

drivers/nec/display/display.cr
pkheav: nice, thanks for that explanation @KimBurgess 

The reason I did 
`      do_send(MsgType::GetParameter, Command::{{name.id}}, priority: priority, name: {{name.id.underscore}})`

and not
`      do_send(MsgType::GetParameter, Command::{{name.id}}, priority: priority, name: {{name.id.underscore.stringify}})`

is because the later actually gives an error in the driver runner even though this evaluates fine in crystal play as well as other alternatives I tried https://play.crystal-lang.org/#/r/9r2l

This is the error that I get:
```In drivers/nec/display/display.cr:325:11

 325 | value.to_s(16, true).rjust(length, '0')
             ^---
Warning: Deprecated Int32#to_s. Use `#to_s(base : Int, *, upcase : Bool = false)` instead

A total of 1 warnings were found.
In lib/placeos-driver/src/placeos-driver.cr:390:1

 390 | macro finished
       ^
Error: expanding macro


There was a problem expanding macro 'finished'

Called macro defined in lib/placeos-driver/src/placeos-driver.cr:390:1

 390 | macro finished

Which expanded to:

    1 |   exec_process_manager = false
    2 | 
    3 |   # Command line options
    4 |   OptionParser.parse(ARGV.dup) do |parser|
    5 |     parser.banner = "Usage: #{PROGRAM_NAME} [arguments]"
    6 | 
    7 |     parser.on("-m", "--metadata", "output driver metadata") do
    8 |       puts Nec::Display::All::KlassExecutor.metadata
    9 |       exit 0
   10 |     end
   11 | 
   12 |     parser.on("-d", "--defaults", "output driver defaults") do
   13 |       puts PlaceOS::Driver::Utilities::Discovery.defaults
   14 |       exit 0
   15 |     end
   16 | 
   17 |     parser.on("-p", "--process", "starts the process manager (expects to have been launched by PlaceOS core)") do
   18 |       exec_process_manager = true
   19 |     end
   20 | 
   21 |     parser.on("-h", "--help", "show this help") do
   22 |       puts parser
   23 |       exit 0
   24 |     end
   25 |   end
   26 | 
   27 |   # Launch the process manager by default, this can be overriten for testing
   28 |   if exec_process_manager
 > 29 |     process = PlaceOS::Driver::ProcessManager.new
   30 | 
   31 |     # Detect ctr-c to shutdown gracefully
   32 |     Signal::INT.trap do |signal|
   33 |       puts " > terminating gracefully"
   34 |       spawn(same_thread: true) { process.terminate }
   35 |       signal.ignore
   36 |     end
   37 | 
   38 |     process.terminated.receive?
   39 |   end
Error: instantiating 'PlaceOS::Driver::ProcessManager.class#new()'


In lib/placeos-driver/src/placeos-driver/process_manager.cr:7:3

 7 | def initialize(@logger_io = ::PlaceOS::Driver.logger_io, @input = STDIN, output = STDERR)
     ^---------
Error: instantiating 'initialize(IO+, IO::FileDescriptor, IO::FileDescriptor)'


In lib/placeos-driver/src/placeos-driver/process_manager.cr:17:43

 17 | @protocol.register :start { |request| start(request) }
                                            ^----
Error: instantiating 'start(PlaceOS::Driver::Protocol::Request)'


In lib/placeos-driver/src/placeos-driver/process_manager.cr:33:3

 33 | def start(request : Protocol::Request, driver_model = nil)
      ^----
Error: instantiating 'start(PlaceOS::Driver::Protocol::Request, Nil)'


In lib/placeos-driver/src/placeos-driver/process_manager.cr:38:28

 38 | driver = DriverManager.new module_id, model, @logger_io, @subscriptions
                             ^--
Error: instantiating 'PlaceOS::Driver::DriverManager.class#new(String, PlaceOS::Driver::DriverModel, IO+, PlaceOS::Driver::Subscriptions)'


In lib/placeos-driver/src/placeos-driver/driver_manager.cr:21:22

 21 | received(data, task)
      ^-------
Error: instantiating 'received(Slice(UInt8), (PlaceOS::Driver::Task | Nil))'


In lib/placeos-driver/src/placeos-driver/driver_manager.cr:207:14

 207 | driver.received(data, task)
              ^-------
Error: instantiating 'Nec::Display::All#received(Slice(UInt8), (PlaceOS::Driver::Task | Nil))'


In drivers/nec/display/display.cr:217:9

 217 | parse_response(ascii_string)
       ^-------------
Error: instantiating 'parse_response(String)'


In drivers/nec/display/display.cr:282:11

 282 | power_on_delay.as(PlaceOS::Driver::Transport)
       ^
Error: can't cast PlaceOS::Driver::Task to PlaceOS::Driver::Transport```

Comment (pkheav) [reply drivers/nec/display/display.cr]: nevermind, ignore this just some bad coding on my end 😞 
https://github.com/PlaceOS/drivers/pull/50/commits/6e81e0cd100effb0f474626ad1a541c53cc1e8a0

File: drivers/nec/display/display.cr (hunk 16 above)
Comment (pkheav) [reply drivers/nec/display/display.cr:297]: I've given this a try but I don't think this is easy to achieve because of the way this protocol is implemented. There is no arbitrary way to know how long data needs to be or how many zeros need to be needed. For example, Command's need to of length 4 which is fine but Inputs need to be of length 2 as well as the state for power if we are setting it ("0004" for on and "0001" for off). Same thing for setting brightness and volume which will be 0x64 or 100 at most but is required to be of length 4.

Oh another case that made this more of a pain is that the Command to Save is required to be of length 2 like this `    do_send(MsgType::Command, "0C", name: "brightness_save")`

File: drivers/nec/display/display.cr (hunk 13 above)
Comment (pkheav) [reply drivers/nec/display/display.cr:162]: I don't think I can do that @KimBurgess as not all of the data that comes back is actually an ascii string and it's a mix of ascii values and hex values https://play.crystal-lang.org/#/r/9sl0

File: drivers/nec/display/display.cr (hunk 16 above)
Comment (pkheav) [reply drivers/nec/display/display.cr]: I've made this commit https://github.com/PlaceOS/drivers/pull/50/commits/d48f502099ee72ee12c300e562fbeedf1151df4f but I'm not sure why I'm getting this error

Comment (pkheav) [reply drivers/nec/display/display.cr]: -> spec calling: power {false}
level=ERROR time=2020-10-05T15:48:28Z source=spec_runner message=executing {"__exec__":"power","power":[false]} on Nec::Display::All (spec_runner)
Expected Int but was Bool at 1:1 (JSON::ParseException)
  from ../usr/share/crystal/src/json/pull_parser.cr:528:5 in 'parse_exception'
  from ../usr/share/crystal/src/json/pull_parser.cr:520:5 in 'expect_kind'
  from ../usr/share/crystal/src/json/pull_parser.cr:126:5 in 'read_int'
  from ../usr/share/crystal/src/json/from_json.cr:76:1 in 'new'
  from ../usr/share/crystal/src/json/from_json.cr:13:3 in 'from_json'
  from drivers/nec/display/display.cr:8:1 in '->'
  from ../usr/share/crystal/src/primitives.cr:255:3 in 'execute'
  from lib/placeos-driver/src/placeos-driver/driver_manager.cr:128:5 in 'execute'
  from lib/placeos-driver/src/placeos-driver/driver_manager.cr:150:18 in 'process'
  from lib/placeos-driver/src/placeos-driver/driver_manager.cr:138:9 in '->'
  from ../usr/share/crystal/src/primitives.cr:255:3 in 'run'
  from ../usr/share/crystal/src/fiber.cr:92:34 in '->'
  from ???

level=ERROR : timeout waiting for expected data
-> expecting: "\u00010*0A0C\u0002C203D60004\u0003\u001D\r"
level=ERROR : unhandled exception in spec
Channel is closed (Channel::ClosedError)
  from ../usr/share/crystal/src/channel.cr:273:20 in 'receive'
  from lib/placeos-driver/src/placeos-driver/driver-specs/runner.cr:433:16 in 'should_send'
  from lib/placeos-driver/src/placeos-driver/driver-specs/runner.cr:415:3 in 'should_send'
  from drivers/nec/display/display_spec.cr:51:4 in '__crystal_main'
  from ../usr/share/crystal/src/crystal/main.cr:105:5 in 'main_user_code'
  from ../usr/share/crystal/src/crystal/main.cr:91:7 in 'main'
  from ../usr/share/crystal/src/crystal/main.cr:114:3 in 'main'
  from ???
... spec failed
... terminating driver gracefully
level=INFO time=2020-10-05T15:48:29Z source=driver.protocol message=IO terminated, exiting cleanly
Driver terminated with: 0
spec runner exited with 0

File: drivers/nec/display/display.cr [hunk 19]
```diff
@@ -0,0 +1,313 @@
...
+    end
+  end
+
+  def disconnected
+    schedule.clear
+  end
+
+  def power(state : Bool)
+    # Do nothing if already in desired state
+    return if self[:power]?.try &.as_bool == state
```
KimBurgess, line 59: You don't need to cast here. Can just use `self[:power]? == state`

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,313 @@
...
+      data = [Command::Power.value, 1]
+      logger.debug { "-- NEC LCD, requested to power on" }
+      do_send(MsgType::Command, data, name: "power", delay: 5.seconds)
+    else
+      logger.debug { "-- NEC LCD, requested to power off" }
+      # 4 = Power Off
+      data = [Command::Power.value, 4]
+      do_send(MsgType::Command, data, name: "power", delay: 10.seconds, timeout: 10.seconds)
+
+      self[:power] = false
```
KimBurgess, line 72: This is likely best set from the parsed response.

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,313 @@
...
+    do_send(MsgType::SetParameter, data, name: "contrast")
+    do_send(MsgType::Command, Command::Save, name: "contrast_save") # Save the settings
+  end
+
+  def volume(val : Int32)
+    data = [Command::VolumeStatus.value, val.clamp(0, 100)]
+
+    do_send(MsgType::SetParameter, data, name: "volume")
+    do_send(MsgType::Command, Command::Save, name: "volume_save") # Save the settings
+    self[:audio_mute] = false # audio is unmuted when the volume is set
```
KimBurgess, line 129: This may also be a good bit of state to update after the device has responded rather than here.

drivers/nec/display/display.cr
KimBurgess: Ah yep, if that's an example of the response there's a few delimiters in there that would need to be used to segment the packet into a few different field first. If the current approach works then stick with that.

drivers/nec/display/display.cr
```diff
@@ -0,0 +1,313 @@
...
+      # auto_setup
+      # nothing needed to do here (we are delaying the next command by 4 seconds)
+    else
+      logger.info { "-- NEC LCD, unknown response: #{data[10..13]}" }
+      logger.info { "-- NEC LCD, full response was: #{data}" }
+    end
+  end
+
+  # Types of messages sent to and from the LCD
+  enum MsgType
```
KimBurgess, line 230: You can specify int subtypes for an enums value. In this case:

```crystal
enum MsgType : UInt8
  ...
end
```

which will remove some of the casting your needing to do with this elsewhere.

KimBurgess (changes requested):
Couple of minor things to neaten up as commented. 

Outside of the driver though it looks like a large amount of changes to other drivers and files in the repo have made their way into this branch / PR. This will need to be cleaned up before merge so that it is introducing changes for the NEC driver only.

pkheav:
Ah damn it, I think messed up the last I tried to rebase this branch from master and now there's 333 commits in this PR but at least now the PR only contains those 2 nec files.

KimBurgess:
Don't stress the number / structure of commits here. We'll squash this down to a single commit for the merge - just important to ensure that squashed commit will only include the intended changes.

Let me know when you're happy with the new rebased state and I'll take a final look.

drivers/nec/display.cr
```diff
@@ -0,0 +1,308 @@
...
+      # Check the checksum equals the second last element
+      logger.debug { "Error: checksum should be #{checksum.to_s(16)}" } unless checksum == data[-2]
+      checksum == data[-2]
+    else
+      true
+    end
+  end
+
+  # Builds the command and creates the checksum
+  private def do_send(type : MsgType, data : Command | Array(Int), **options)
```
KimBurgess, line 277: There's a couple of things going on in here where it may be good to take a slightly different approach and neaten up.

The passing of the message type and an accompanying payload is nice and looks to match the structure of the device protocol well. Let look at the payload itself though - the only common abstraction here between _all_ message types is it is just a list of bytes. With each message type though there is a common structure so rather than handling everything here, this method could take a `MsgType` and a `data : Bytes`, then purely focus on wrapping these into the full command packet (forming the heading, adding the framing to the message, checksumming etc).

Upstream on this you could then have three methods `set_parameter`, `get_parameter` and `exec_command` which take args as appropriate and purely focus on forming up the `data` that's passed here.

drivers/nec/display.cr
```diff
@@ -0,0 +1,308 @@
...
+    data = [data.value] if data.is_a?(Command)
+    data = data.join { |i| format_value(i) }.bytes
+    bytes = Bytes.new(data.size + 11)
```
KimBurgess, line 280: When it comes to building the packet here, take a look at [String.build](https://crystal-lang.org/api/0.35.1/String.html#build(capacity=64,&):self-class-method). There's a tiny bit of overhead here that just using `Bytes` directly, but this may be counteracted by the improved readability.

This will make concat'ing multiple byte values, and some of the type conversion a little clear:
```crystal
String.build do |str|
  str.write_byte 0x01 # write bytes directly
  str << "hello"      # or strings
  str << 'c'          # and char values too
  str << 1            # integers will write as their direct string form
end
```

drivers/nec/display.cr
KimBurgess: If you go this route, it should help neaten up some of the magic nums that are still in use for the power control.

KimBurgess (commented):
Couple of minor notes that may help with cleanup. These are entirely optional and more to present / discuss different methods for approaching these implementations. If this is functional we can merge in the current state, then continue to optimise as time allows.

drivers/nec/display.cr
pkheav: Thanks @KimBurgess , I've made this pretty big commit to address your latest 2 comments.

So I think the protocol is general enough so that all commands will be in this general format:
`#{msg_type}#{command}#{data}`

`data` is not required for all commands so is only passed in when necessary which is for any commands with `MsgType::SetParameter` and in the only exception for `def power` for commands with `MsgType::Command`

pkheav: Also, I couldn't figure out how to generate the checksum with one `String.build` block annoyingly so I had to settle on doing this.

https://github.com/PlaceOS/drivers/blob/10db226249ede7583b834fb73a52b839a4f5004e/drivers/nec/display.cr#L285-L301

Not sure if it's possible to generate the entire command string in one `String.build` block unless you know some way to do this @KimBurgess

pkheav: Oh just realised I had to use `String.build` 3 times actually :/

drivers/nec/display.cr
```diff
@@ -0,0 +1,303 @@
...
+    PowerOnDelay     = 0x02D8
+    ContrastStatus   = 0x0012
+    BrightnessStatus = 0x0010
+    AutoSetup        = 0x001E
+    PowerQuery       = 0x01D6
+    Save             = 0x0C
+    SetPower         = 0xC203D6
+  end
+
+  private def format_value(value : Int, length : Int = 4) : String
```
KimBurgess, line 235: This method doesn't look to be used after the refactor.

drivers/nec/display.cr
```diff
@@ -0,0 +1,303 @@
...
+    def build(command : Nec::Display::Command, data : Int? = nil)
+      message = format_value(command.value)
+
+      message = String.build do |str|
+        str << "0*0"
+        str.write_byte self.value                # Type
+
+        message_length = message.size + 2
+        message_length += 4 if d = data          # If there is data, add 4 to the message length
+        str << format_value(message_length, 2)   # Message length
```
KimBurgess, line 286: Now that some of this formatting is neater to inline the conversion in place of a seperate method call.

drivers/nec/display.cr
```diff
@@ -0,0 +1,303 @@
...
+
+    def format_value(value : Int, length : Int = 4) : String
+      length = 2 if value == 0x0C # To deal with Command::Save
+      value.to_s(16, true).rjust(length, '0')
+    end
+
+    def build(command : Nec::Display::Command, data : Int? = nil)
+      message = format_value(command.value)
+
+      message = String.build do |str|
```
KimBurgess, line 280: Avoid overriding the same variable for multiple uses.

drivers/nec/display.cr
```diff
@@ -0,0 +1,303 @@
...
+        str << message                           # Message
+        str << format_value(d) if d              # Data if required
+        str.write_byte 0x03                      # End of message
+      end
+
+      String.build do |str|
+        str.write_byte 0x01                      # SOH
+        str << message
+        checksum = 0x00_u8
+        message.each_byte { |b| checksum = checksum ^ b }
```
KimBurgess, line 297: You can use `#reduce` to do this directly without external assignment. Either way works, purely commenting as an alternative style that can make these operations very concise.

```crystal
message.each_byte.reduce { |a, b| a ^ b }
```

drivers/nec/display.cr
KimBurgess: Nice. Heaps neater.

drivers/nec/display.cr
pkheav: Thanks @KimBurgess , yeah that's way neater

drivers/nec/display.cr
```diff
@@ -0,0 +1,299 @@
...
+
+  def auto_adjust
+    data = MsgType::SetParameter.build(Command::AutoSetup, 1)
+    send(data, name: "auto_adjust")
+  end
+
+  def brightness(val : Int32)
+    data = MsgType::SetParameter.build(Command::BrightnessStatus, val.clamp(0, 100))
+    send(data, name: "brightness")
+    send(MsgType::Command.build(Command::Save), name: "brightness_save") # Save the settings
```
KimBurgess, line 109: The save command looks to be the same regardless of what state you're trying to persist. With this in mind, to remove redundancy they can all be named the same and set to a low priority. That way if a set of settings are applied at once, `Command::Save` will only be sent once after all of these have been applied.

If you do this, just make sure the the result from the `send` for the actual setting change is returned, otherwise callers of the methods will be returned an error if their save command is overwritten (which is desired).

drivers/nec/display.cr
```diff
@@ -0,0 +1,299 @@
...
+    end
+
+    logger.debug { "NEC LCD responded with ascii_string #{ascii_string}" }
+
+    command = MsgType.from_value(data[4])
+
+    case command # Check the MsgType (B, D or F)
+    when .command_reply?
+      # Power on and off
+      if ascii_string[10..15] == "C203D6" # Means power comamnd
```
KimBurgess, line 159: With some of the restructuring, these magic values can now likely be replaced by what's in the `Command` enum.

drivers/nec/display.cr
```diff
@@ -0,0 +1,299 @@
...
+  {% for name in Command.constants %}
+    @[Security(Level::Administrator)]
+    def {{name.id.underscore}}(priority : Int32 = 0)
+      send(MsgType::GetParameter.build(Command::{{name.id}}), priority: priority, name: {{name.id.underscore.stringify}})
+    end
+  {% end %}
+
+  private def check_checksum(data : Bytes)
+    # Loop through the second to the third last element
+    if data.size >= 2
```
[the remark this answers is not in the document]
pkheav: Looking at the docs, I don't think so so I've removed this which I just copied over from the old ruby driver

https://github.com/PlaceOS/drivers/pull/50/commits/a0976b703d39ed7a5d0d1843d7afd6adfbfce525

drivers/nec/display.cr
pkheav: I've made this commit but this still remains as the power query response is an annoyingly unique case.

https://github.com/PlaceOS/drivers/pull/50/commits/5d61fd20eddf243f99cc26bbe653a31b5181ee0f#diff-7c4dd3c6e0b44d710bc595343e7b26c65e527df44935f29f97927f6bd22c0904R153-R155

image

drivers/nec/display.cr
```diff
@@ -0,0 +1,281 @@
...
+    # Check for valid response
+    if !check_checksum(data)
+      return task.try &.retry("-- NEC LCD, invalid response was: #{ascii_string}")
+    end
+
+    logger.debug { "NEC LCD responded with ascii_string #{ascii_string}" }
+
+    if ascii_string[8..9] == "00"
+      parse_response(ascii_string)
+    elsif ascii_string[10..13] == "00D6" # Annoyingly unique case to deal with power status query
```
KimBurgess, line 153: Rather than splitting on this, take a look at the 5th byte of the header. This is the "message type" which you can parse into the `MsgType` Enum already defined.

Power query does indeed look to need some special handling, but this should make it clearer that's it's due to a command response vs some of other message types that share a common structure. 
9DBD913D-EF8E-48EE-8139-D8C2724C68EC

drivers/nec/display.cr
```diff
@@ -0,0 +1,281 @@
...
+      volume_status
+      video_input
+      audio_input
+    end
+  end
+
+  # LCD Response code
+  def received(data, task)
+    task_name = task.try &.name
+    logger.debug { "task is #{task_name}" }
```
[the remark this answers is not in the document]
pkheav: ah, thanks Kim, I missed that when cleaning up the driver. I've removed it as it was only for debugging.

drivers/nec/display.cr
pkheav: I've made this commit which hopefully makes things clearer https://github.com/PlaceOS/drivers/pull/50/commits/e86303c89c2085c72a62c07d1cde45142d087397

KimBurgess: [You are still branching](https://github.com/PlaceOS/drivers/commit/e86303c89c2085c72a62c07d1cde45142d087397#diff-7c4dd3c6e0b44d710bc595343e7b26c65e527df44935f29f97927f6bd22c0904R165) on a couple of bytes that are specific to the power query response. Based on the protocol design there's no guarantee these are unique only to that bit of comms. You'll need to first check the response type (there's a header byte for this), then _if_ its a command reply check the command type and if that's a power query command grab out the value.

pkheav: Ah I see what you mean. I think this should address that

https://github.com/PlaceOS/drivers/pull/50/commits/a890cbad481eb23abb26d67e2bb78b3f373740db

drivers/nec/display.cr
```diff
@@ -0,0 +1,280 @@
...
+      self[:volume] = 0 if value == 1
+    when .auto_setup?
+      # auto_setup
+      # nothing needed to do here (we are delaying the next command by 4 seconds)
+    when .set_power?
+      self[:power] = value == 1
+    else
+      logger.debug { "-- NEC LCD, unknown response received: #{data}" }
+    end
+  end
```
KimBurgess, line 204:
```suggestion
  def received(data, task)
    header = data[0..6]
    message = data[7..-3]
    checksum = data[-2]

    unless checksum == data[1..-3].reduce { |a, b| a ^ b }
      return task.try &.retry("invalid checksum in device response")
    end

    begin
      case MsgType.from_value header[4]
      when .command_reply?
        parse_command_reply message
      when .get_parameter_reply?, .set_parameter_reply?
        parse_response message
      else
        raise "unknown message type"
      end
    rescue e
      task.try &.abort e.message
    else
      task.try &.success
    end
  end

  # Command replies each use a different packet structure
  private def parse_command_reply(message : Bytes)
    response = String.new(message[1..-2]).hexbytes

    if response[1..3] == Bytes[0xC2, 0x03, 0xD6] # Set power
      result_code = response[0]
      raise "unsupported operation" unless result_code == 0
      self[:power] = response[5] == 1
    elsif response[2..3] == Bytes[0xD6, 0x00] # Power query
      result_code = response[1]
      raise "unsupported operation" unless result_code == 0
      self[:power] = response[7] == 1
    else
      logger.warn { "unhandled command reply: #{message}" }
    end
  end

  # Get and set parameter replies share common structure
  private def parse_response(message : Bytes)
    response = String.new(message[1..-2]).hexbytes

    result_code = response[0]
    raise "unsupported operation" unless result_code == 0

    op_code = response[1].to_u16 << 8 | response[2]
    value = response[6].to_u16 << 8 | response[7]

    case Command.from_value op_code
    when .video_input?
      self[:input] = Input.from_value(value)
    when .audio_input?
      self[:audio] = Audio.from_value(value)
    when .volume_status?
      self[:volume] = value
      self[:audio_mute] = value == 0
    when .brightness_status?
      self[:brightness] = value
    when .contrast_status?
      self[:contrast] = value
    when .mute_status?
      self[:audio_mute] = value == 1
      self[:volume] = 0 if value == 1
    when .auto_setup?
      # auto_setup
      # nothing needed to do here (we are delaying the njxt command by 4 seconds)
    else
      logger.warn { "unhandled device response: #{message}" }
    end
  end
```

KimBurgess (commented):
Popped a quick refactor inline for the response parsing - this is just to seperate those weird command reply packets and remove some of the cruft from previous iterations.

Aside from that, can you drop all of the `-- NEC LED,...` prefixes from log messages. These will always appear with the module / driver details as context so that'll remove some noise.

Tests looks good. Once the above two items are resolved, good for merge.

pkheav:
> Popped a quick refactor inline for the response parsing - this is just to seperate those weird command reply packets and remove some of the cruft from previous iterations.
> 
> Aside from that, can you drop all of the `-- NEC LED,...` prefixes from log messages. These will always appear with the module / driver details as context so that'll remove some noise.
> 
> Tests looks good. Once the above two items are resolved, good for merge.

Just made this commit as well to handle the save response which wasn't been handled in the refactor https://github.com/PlaceOS/drivers/pull/50/commits/2ec673e83a02cdccfae6e94132b957d8ba6ef228 and this commit to cleanup the debugging messages/not used code https://github.com/PlaceOS/drivers/pull/50/commits/124d6edfc5312f74ca449e6c4ab98141cd2905ef

KimBurgess (approved):
LGTM. Let's squash this monster of PR and get it merged as an initial version.

Notes for additional bits of refactoring if there's future work on this driver:
1. Prune the `Command` enum to only be `Parameters` used by the get/set commands. This will allow removal of the branching in `#to_s`.
2. Refactor TX string building to seperate message body construction and the wrapping with header, check code and delim. This will provide a cleaner implementation for get/set commands (which share structure) and allow custom message bodies to be inserted in the wildly varying command type messages.
3. Use a [callback based response handler](https://github.com/PlaceOS/drivers/blob/f5884166624be803cc2864daa0a50567f1b77944/docs/writing-a-driver.md#streaming-io) for command style messages as their response format is specific to the command.

PR #61, opened 2020-10-08
Lg lcd

pkheav:


drivers/lg/displays/ls5.cr
```diff
@@ -0,0 +1,291 @@
...
+        info
+      }
+    else
+      logger.debug { "No MAC address provided" }
+    end
+  end
+
+  DUPLICATES = ["Contrast", "NoSignalOff", "AutoOff", "PmMode"]
+  def received(data, task)
+    command = Command.from_value(data[0]).to_s
```
KimBurgess, line 221: It will be much safer to work with this as it's enum value rather than in string form.

drivers/lg/displays/ls5.cr
```diff
@@ -0,0 +1,291 @@
...
+    when "LocalButtonLock"
+      self[:local_button_lock] = resp_value == 2
+    else
+      return task.try &.retry
+    end
+
+    task.try &.success
+  end
+
+  private def do_send(command : Command, data : Int, system : Char = 'k', **options)
```
KimBurgess, line 281: What special about system `k`? Can this be given some meaning from either encoding this in a value, or at a minimum referencing in a comment?

drivers/lg/displays/ls5.cr
```diff
@@ -0,0 +1,291 @@
...
+    # Special case for PM Mode
+    if command.pm_mode? && system == 's'
```
KimBurgess, line 283: Ditto here for `s`.

drivers/lg/displays/ls5_spec.cr
```diff
@@ -0,0 +1,64 @@
...
+  exec(:power, false)
+  # mute_video(true)
+  should_send("kd 01 01\r")
+  responds("d 01 OK01x")
+  status[:power].should eq(false)
+  # mute_audio(true)
+  should_send("ke 01 00\r")
+  responds("e 01 OK00x")
+  status[:audio_mute].should eq(true)
+end
```
KimBurgess, line 64: Would be good to get some more specs here at least for input switching.

drivers/lg/displays/ls5.cr
pkheav: https://github.com/PlaceOS/drivers/blob/3dea9fef83f703d0af9ef10443dda86bf851bfd9/drivers/lg/displays/ls5.cr#L286-L289

drivers/lg/displays/ls5.cr
pkheav: Yeah I tried using the enum values for this but because some of the commands use the same letter/value, then I can't actually distinguish between their responses alone from the first character in the response string.

Pretty much if there's duplicate values in an enum, it'll always default to the one defined first so AutoOff in this case. Let me know if you have a better idea of dealing with this https://play.crystal-lang.org/#/r/9tt5

Also, this comments try to explain that https://github.com/PlaceOS/drivers/blob/3dea9fef83f703d0af9ef10443dda86bf851bfd9/drivers/lg/displays/ls5.cr#L223-L225

drivers/lg/displays/ls5_spec.cr
pkheav: https://github.com/PlaceOS/drivers/pull/61/commits/97bb47f899df9115d3e10bbec864e350f60d9cb0

drivers/lg/displays/ls5.cr
KimBurgess: Ah I see. Looking at the protocol, commands are two bytes. Rather than fudging things on the parsing, it be more reliable to rework the way this is modelled to capture this.

One approach that could work is to pack the two command bytes (`UInt8`'s) into a single `UInt16`. This will then let the commands be represented, without ambiguity and also ensure that everything needed to serialise and deserialise them is define clearly in one spot.

```crystal
enum Command : UInt16
  Power    = 0x6b61 # k a
  SetInput = 0x7962 # x b
  # etc
  
  def self.new(data1, data2)
    self.new((data1.to_u16 << 8) ^ data2.to_u8)
  end
end

p! Command.new('k'.ord, 'a'.ord)
```
https://play.crystal-lang.org/#/r/9tvu

The only slightly annoying thing is I'm not aware of a way to get `Char#ord` values from within a macro or force compile time eval of these to make the actual value declarations a little neater. E.g. ideally this would be:

```crystal
enum Command : UInt16
  Power = bitpack('k', 'a')
  # ...
end
```

Tagging @stakach and @caspiano for other thoughts on this one too as defining a clean way to express multi-byte Enum's is a common problem worth solving.

KimBurgess: P.S. see crystal-lang/crystal#9830 for a proposal to solve this at the language level. Hopefully it's generally useful, however if there's no interest there can look at adding something to the driver framework.

pkheav: Hey @KimBurgess , if I do switch over to packing 2 bytes into a command like `NoSignalOff    = 0x6667 # fg`, I'm not sure how I'll be able to use the enum to parse responses.

So the command for `no_signal_off(false)` is `"fg 01 00\r"` and the response is `"g 01 OK00x"`. Since Command::NoSignalOff is representative of the two bytes which are only present in the command that's sent and the response only contains "g" and not "fg", this will make it difficult if not impossible to figure out what command the response is actually from.

KimBurgess: Yep - I see what you mean. In cases like this where the protocol does not provide the ability to uniquely distinguish the response from the response data alone, you may want to define response parsing as a callback block, in place of the single `received` method.

Take a look at ["send and callback" here](https://github.com/PlaceOS/drivers/blob/master/docs/writing-a-driver.md#streaming-io).

drivers/lg/displays/ls5.cr
```diff
@@ -0,0 +1,297 @@
...
+        wake(broadcast || @last_broadcast)
+      end
+    end
+    # To power on, unmute the display
+    # To power off, mute the display
+    mute(!state) if self[:connected]
+  end
+
+  def hard_off
+    do_send(Command::Power, 0, name: "power", priority: 99)
```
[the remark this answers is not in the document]
pkheav: ah okay, thanks, that's good to know

drivers/lg/displays/ls5.cr
```diff
@@ -0,0 +1,297 @@
...
+      if self[:hard_power]?.try &.as_bool
+        screen_mute?
+        input?
+        volume_mute?
+        volume?
+      end
+    elsif self[:connected]?.try &.as_bool
+      screen_mute?
+
+      if @id_num == 1
```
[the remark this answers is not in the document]
pkheav: No idea why actually as I've copied this over from the ruby driver. I can get rid of this but it might have been there for a reason?

Not exactly of the difference between communicating with a device that's hooked up via rs232 compared to one that's not regarding this bit of code as well which I also carried over from the ruby driver
https://github.com/PlaceOS/drivers/blob/19d0468789bbed049f8f42aa2d84de565404a631/drivers/lg/displays/ls5.cr#L145-L160

pkheav:
Ah, I did not know about the send and callback which yeah is useful for cases like this. Thanks @KimBurgess , I've made this commit https://github.com/PlaceOS/drivers/pull/61/commits/c3131c421e63d9e813170f6379f6cad41b5a39eb

drivers/lg/displays/ls5.cr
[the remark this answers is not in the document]
stakach: I think from a Crystal engine point of view, `self[:connected] == JSON::Any.new(true / false)`
So `JSON::Any.new(false)` evaluates to `true` so need to use `.raw` or move it into the expected type via `as_bool` (will raise an error if an Int etc)

The real answer is to track `@connected` state via an instance variable and just update the public state as required (much more performant too)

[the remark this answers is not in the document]
pkheav: Thanks @stakach , I've made these commits.

https://github.com/PlaceOS/drivers/pull/61/commits/d5c37cbd170c895a8be5cbff624ee28661b3758f

https://github.com/PlaceOS/drivers/pull/61/commits/2d7d4fc2158a4f93510527dac15f827f52a6b1c7

PR #62, opened 2020-10-09
Epson projector

### LINKED ISSUE #29: Epson Projector

dukeraphaelng:
This PR will:
- Resolve #29

### DIFF
--- a/drivers/epson/projector/esc_vp21.cr
+++ b/drivers/epson/projector/esc_vp21.cr
@@ -39,40 +39,45 @@ class Epson::Projector::EscVp21 < PlaceOS::Driver
     if state
       @power_target = true
       logger.debug { "-- epson Proj, requested to power on" }
-      do_send("PWR", "ON", delay: 40.seconds, name: "power")
+      do_send(:power, "ON", delay: 40.seconds, name: "power")
     else
       @power_target = false
       logger.debug { "-- epson Proj, requested to power off" }
-      do_send("PWR", "OFF", delay: 10.seconds, name: "power")
+      do_send(:power, "OFF", delay: 10.seconds, name: "power")
     end
     power?
   end
 
   def power?(**options) : Bool
-    do_send("PWR", **options, name: :power?).get
+    do_send(:power, **options, name: :power?).get
     !!self[:power]?.try(&.as_bool)
   end
 
   def switch_to(input : Input)
     logger.debug { "-- epson Proj, requested to switch to: #{input}" }
-    do_send("SOURCE", input.value.to_s(16), name: :input)
+    do_send(:input, input.value.to_s(16), name: :input)
     self[:input] = input # for a responsive UI
     self[:video_mute] = false
     input?
+  end
+
+  def input?
+    do_send(:input, name: :input_query, priority: 0).get
+    self[:input]
   end
 
   # Volume commands are sent using the inpt command
   def volume(vol : Int32, **options)
     vol = vol.clamp(0, 255)
     self[:unmute_volume] = self[:volume] if mute = vol == 0 # Store the "pre mute" volume, so it can be restored on unmute
-    do_send("VOL", vol, **options, name: :volume)
+    do_send(:volume, vol, **options, name: :volume)
     self[:volume] = vol
     self[:audio_mute] = mute
     volume?
   end
 
   def volume?
-    do_send("VOL", name: :volume?, priority: 0).get
+    do_send(:volume, name: :volume?, priority: 0).get
     self[:volume]?.try(&.as_i)
   end
 
@@ -83,8 +88,8 @@ class Epson::Projector::EscVp21 < PlaceOS::Driver
   )
     case layer
     when .audio_video?
-      do_send("MUTE", state ? "ON" : "OFF", name: :mute)
-      do_send("MUTE", name: :mute?, priority: 0)
+      do_send(:av_mute, state ? "ON" : "OFF", name: :mute)
+      do_send(:av_mute, name: :mute?, priority: 0)
     when .video?
       mute_video(state)
     when .audio?
@@ -93,7 +98,7 @@ class Epson::Projector::EscVp21 < PlaceOS::Driver
   end
 
   def mute_video(state : Bool)
-    do_send("MSEL", state ? "ON" : "OFF", name: :video_mute)
+    do_send(:video_mute, state ? "ON" : "OFF", name: :video_mute)
     video_mute?
   end
 
@@ -106,13 +111,8 @@ class Epson::Projector::EscVp21 < PlaceOS::Driver
   end
 
   def video_mute?
-    do_send("MSEL", name: :video_mute?, priority: 0).get
+    do_send(:video_mute, name: :video_mute?, priority: 0).get
     !!self[:video_mute]?.try(&.as_bool)
-  end
-
-  def input?
-    do_send("SOURCE", name: :input_query, priority: 0).get
-    self[:input]
   end
 
   ERRORS = [
@@ -139,14 +139,29 @@ class Epson::Projector::EscVp21 < PlaceOS::Driver
     "16: IF board discernment error"
   ]
 
+  def inspect_error
+    do_send(:error, priority: 0)
+  end
+
+  COMMAND = {
+    power: "PWR",
+    input: "SOURCE",
+    volume: "VOL",
+    av_mute: "MUTE",
+    video_mute: "MSEL",
+    error: "ERR",
+    lamp: "LAMP"
+  }
+  RESPONSE = COMMAND.to_h.invert
+
   def received(data, task)
     return task.try(&.success) if data.size <= 2
     data = String.new(data[1..-2])
     logger.debug { "epson Proj sent: #{data}" }
 
     data = data.split('=')
-    case data[0]
-    when "ERR"
+    case RESPONSE[data[0]]
+    when :error
       if data[1]?
         code = data[1].to_i(16)
         self[:last_error] = ERRORS[code]? || "#{data[1]}: unknown error code #{code}"
@@ -154,7 +169,7 @@ class Epson::Projector::EscVp21 < PlaceOS::Driver
       else # Lookup error!
         return task.try(&.abort("Epson PJ sent error response for #{task.not_nil!.name || "unknown"}"))
       end
-    when "PWR"
+    when :power
       state = data[1].to_i
       self[:power] = state < 3
       self[:warming] = state == 2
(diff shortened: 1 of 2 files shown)


drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+require "placeos-driver/interface/powerable"
+require "placeos-driver/interface/switchable"
+
+module Epson; end
+
+module Epson::Projector; end
+
+# Documentation: documentation link
+
+class Epson::Projector::EscVp21 < PlaceOS::Driver
```
KimBurgess, line 12: I'm not familiar with the protocol for this device, but unless this is specific to an exact model, can probably make this class make a little more generic.

drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+  include Interface::Muteable
+
+  enum Inputs
+    HDMI
+    HDBaseT
+  end
+
+  # Discovery Information
+  tcp_port 1024
+  descriptive_name "Epson Projectors"
```
KimBurgess, line 23:
```suggestion
  descriptive_name "Epson Projector"
```

drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+  generic_name :Display
+
+  def on_load
+    transport.tokenizer = Tokenizer.new("\r\n")
+
+    self[:volume_min] = 0
+    self[:volume_max] = 255
+
+    self[:power] = 0
```
KimBurgess, line 32: Best to set any initial state based on a device query, rather than explicit value.

drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+  end
+
+  def on_update
+  end
+
+  def connected
+    # Have to init comms
+    send("ESC/VP.net\x10\x03\x00\x00\x00\x00")
+    do_poll
+    schedule.every(52.seconds) { do_poll }
```
KimBurgess, line 45: In cases where you want to create a repeating schedule and have it run straight away too you can use `schedule.every(52.seconds, true) { ... }`

drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+  def power?(**options, &block)
+    options[:emit] = block unless block.nil?
+    options[:name] = :power_state
+    do_send(:PWR, **options)
+  end
+
+  #
+  # Input selection
+  #
+  INPUTS = {
```
KimBurgess, line 85: You can encode these values within the existing Inputs enum.

drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+  }
+  INPUT_LOOKUP = INPUTS.invert
+
+  def switch_to(input : Inputs)
+    do_send(:SOURCE, INPUT_LOOKUP[input], name: :inpt_source)
+    do_send(:SOURCE, name: :inpt_query)
+
+    logger.debug { "-- epson LCD, requested to switch to: #{input}" }
+    self[:input] = input # for a responsive UI
+    self[:MUTE] = false
```
KimBurgess, line 97:
```suggestion
    self[:mute] = false
```

drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+  end
+
+  #
+  # Volume commands are sent using the inpt command
+  #
+  def volume(vol : Int32, **options)
+    vol = 0 if vol < 0
+    vol = 255 if vol > 255
```
KimBurgess, line 105: Take a look at [clamp](https://crystal-lang.org/api/0.35.1/Comparable.html#clamp(min,max)-instance-method) for this use case.

drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+
+    # Seems to only return ":" for this command
+    self[:volume] = vol
+    self[:unmute_volume] = vol if vol > 0 # Store the "pre mute" volume, so it can be restored on unmute
```
KimBurgess, line 109: This can probably be stored as an internal variable rather than exposed state.

drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+
+  # Mutes audio + video
+  def mute(
+    state : Bool = true,
+    index : Int32 | String = 0,
+    layer : MuteLayer = MuteLayer::AudioVideo
+  )
+    logger.debug { "-- epson Proj, requested mute state: #{state}" }
+
+    do_send(:MUTE, state, name: :video_mute) # Audio + Video
```
KimBurgess, line 121: You need to honour the `layer` param here.

drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+
+  def unmute_audio(index : Int32 | String = 0)
+    mute_audio(false, index)
+  end
+
+  def input?
+    do_send(:SOURCE, name: :inpt_query, priority: 0)
+  end
+
+  ERRORS = {
```
KimBurgess, line 145: This looks like a good spot for an enum.

drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+    19 => "13: power supply unit error",
+    20 => "14: exhaust shutter error",
+    21 => "15: obstacle detection error",
+    22 => "16: IF board discernment error",
+  }
+
+  #
+  # epson Response code
+  #
+  def received(data, resolve, command) # Data is default received as a string
```
KimBurgess, line 172: This method signature looks a bit off - the received callback is passed the data and a task object.

drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+    logger.debug { "epson Proj sent: #{data}" }
+
+    if data == ":"
+      return :success
```
KimBurgess, line 176: Have a read here for now to mark a task result: https://github.com/PlaceOS/drivers/blob/master/docs/writing-a-driver.md#queue

drivers/epson/projector/esc_vp21.cr
```diff
@@ -0,0 +1,274 @@
...
+    end
+
+    data = data.split(/=|\r:/)
+    case data[0].to_sym
```
KimBurgess, line 180: `to_sym` is a Ruby method. This will need some crystallising.

KimBurgess (changes requested):
I've added a couple of initial comments - looks like there's still a little way to go but more than happy to keep some conversation happening in-line here as you have queries.

One thing I will recommend for both this driver and any others is that the focus should not be on porting them across verbatim. The old Ruby drivers may act as a good guide to highlight any device oddities, but not something to reproduce. Crystal is a language of its own and there's a lot of opportunity to build these neatly and in a way that uses the best of what crystals type system has to offer. Spend some time to grok how the old driver is implemented, then use this knowledge to help model the new driver in a way that makes sense for the tools at hand, yourself and others working on it in the future.

drivers/epson/projector/esc_vp21.cr
dukeraphaelng: Sorry what do you mean by 'honour'?

drivers/epson/projector/esc_vp21.cr
dukeraphaelng: Could you clarify what you mean by a 'device query'?

KimBurgess: Sending a command to the device that returns the device state. For example, it looks like you already have `power?` and `input?` implemented for this.

The reason for this is if a driver is stopped from backoffice, when it restarts you don't want to generate state change events _unless_ the state has actually changed. A nice pattern to follow is use the `connected` method to query any state that the driver is exposing. That way if the device drops offline or the driver stops when connectivity is re-established the driver always shows the correct device state.

drivers/epson/projector/esc_vp21.cr
KimBurgess: At the moment if `mute(true, layer: MuteLayer::Audio)` is called, this method will activate the device video mute. You just need to check that param, then execute the appropriate action(s).

dukeraphaelng:
> Having a quick read now. To provide some further reading on `**options`, that's a use of [splatting](https://crystal-lang.org/reference/syntax_and_semantics/splats_and_tuples.html).
> 
> TL;DR it can be used as a continent way to forward an arbitrary set of keyword arguments around. It was used a lot in the previous Ruby drivers, but would recommend avoiding it unless absolutely needed here as it can lead to some sloppy type signatures.

Hi @KimBurgess , upon going through this comment In Gab's PR I thought it'd be a good idea to implement `Hash` instead of `splat`. However, this is not allowed since `splatting` in drivers is mostly used in 

```
queue(**({
      name: command,
    }.merge(options))) do
...
end
```

and queue's method is defined in `placeos-driver/src/placeos-driver.cr` as follows (line 114-117)
```
  # Queuing
  def queue(**opts, &block : Task -> Nil)
    @__queue__.add(**opts, &block)
  end
```
which disallows anything but splats and NamedTuple in the argument.

KimBurgess:
That `options` forwarding can stay out of these initial driver implementation, then can add this down the track when / where appropriate.

For some further info, you can't splat a hash in Crystal. You could in Ruby, but this breaks when we introduce types and the safety they provide. Within crystal you can think of a double splat as some syntactic sugar.

Say we have the following method:
```crystal
def foo(a : String, b : Int32)
  # ...
end
```
We can call this via
```crystal
foo a: "hello", b: 42
```
But if we also had those params ready to go somewhere else in a NameTuple, the following is equivalent.
```crystal
x = {a: "hello", b: 42}
foo **x
```

The also works in the opposite direction
```crystal
def bar(**kwargs)
  p! kwargs
end

bar a: "hello", b: 42
#=> {a: "hello", b: 42}
```

This is the same for `*var`, which expands or captures `Tuples` to/from positional params too.

drivers/epson/projector/esc_vp21.cr
pkheav: @dukeraphaelng Something like this https://github.com/PlaceOS/drivers/blob/master/drivers/samsung/displays/mdc_protocol.cr#L124-L132

or for the epson projector example
https://github.com/PlaceOS/drivers/pull/62/commits/60e64c38c5e7d7e967e94d098db0356ba3c55dd7#diff-a9b9e76ba00d89e0a1d6dcd683831bc983f1e69e2ac6736f710d2f4909bb3ca3R110-R116

drivers/epson/projector/esc_vp21.cr
pkheav: I think ESC/VP21 is just the name of the protocol or maybe group of projectors 
image

PR #65, opened 2020-10-14
feat: Planar Clarity Video Wall

### LINKED ISSUE #59: Planar Clarity Video Wall

GabFitzgerald:
First attempt
Resolves #59 

some confusion with options / **options


KimBurgess:
Having a quick read now. To provide some further reading on `**options`, that's a use of [splatting](https://crystal-lang.org/reference/syntax_and_semantics/splats_and_tuples.html).

TL;DR it can be used as a continent way to forward an arbitrary set of keyword arguments around. It was used a lot in the previous Ruby drivers, but would recommend avoiding it unless absolutely needed here as it can lead to some sloppy type signatures.

drivers/planar/clarity_matrix.cr
```diff
@@ -0,0 +1,111 @@
...
+
+  #   implements :device # don't need?
+
+  def on_load
+    # Communication settings
+    queue.wait = false
+    transport.tokenizer = Tokenizer.new("\r")
+  end
+
+  def on_unload
```
KimBurgess, line 23: When writing drivers, if there aren't any actions that you need to run in these lifecycle callback methods, you can omit them.

drivers/planar/clarity_matrix.cr
```diff
@@ -0,0 +1,111 @@
...
+  end
+
+  def on_update
+  end
+
+  def connected
+    schedule.every(60.seconds) { do_poll }
+    do_poll
```
KimBurgess, line 31: `schedule.every` takes an optional second arg. It this is `true` it will run the scheduled block straight away in addition to the schedule so that you don't need to repeat like this.

drivers/planar/clarity_matrix.cr
```diff
@@ -0,0 +1,111 @@
...
+
+    # options[:wait] = true
+    # options[:name] = :pwr_query
+    send("op A1 display.power ? \r", **options)
+  end
+
+  # def power(state, broadcast_ip = false, options = {})
+  # what does broadcast_ip do here??
+  def power(state : Bool = false)
+    puts "power method running"
```
KimBurgess, line 54: Rather than using ours, take a look at the [logger](https://github.com/PlaceOS/drivers/blob/master/docs/writing-a-driver.md#logger).

drivers/planar/clarity_matrix.cr
```diff
@@ -0,0 +1,111 @@
...
+
+  # def power(state, broadcast_ip = false, options = {})
+  # what does broadcast_ip do here??
+  def power(state : Bool = false)
+    puts "power method running" 
+    # self[:power] = state
+    # send("op A1 display.power = off \r")
+    if state == true
+      send("op A1 display.power = on \r") # changed ** -> A1: "Power should only be sent to display A1"
+      power?
```
KimBurgess, line 59: It looks like the power state gets updated in remove to the power on command above, this second query can be removed so there's not unnecessary comms.

drivers/planar/clarity_matrix_spec.cr
```diff
@@ -0,0 +1,10 @@
+DriverSpecs.mock_driver "Planar::ClarityMatrix" do
+  # on connect it should do_poll the device
+  should_send("op A1 display.power ? \r")
```
KimBurgess, line 3: When writing specs, you also want to mock mock the device responses so that you test full request -> response behaviour. For some info on this, see: https://github.com/PlaceOS/drivers/blob/master/docs/writing-a-spec.md#testing-streaming-io.

drivers/planar/codes.txt
```diff
@@ -0,0 +1,17 @@
+op ** display.power = on             // Power ON **check this
```
KimBurgess, line 1: Try to keep any notes like this, or other items not specifically relating to the driver out of this repo.

dukeraphaelng:
> Having a quick read now. To provide some further reading on `**options`, that's a use of [splatting](https://crystal-lang.org/reference/syntax_and_semantics/splats_and_tuples.html).
> 
> TL;DR it can be used as a continent way to forward an arbitrary set of keyword arguments around. It was used a lot in the previous Ruby drivers, but would recommend avoiding it unless absolutely needed here as it can lead to some sloppy type signatures.

Hi Kim, upon going through this comment I thought it'd be a good idea to implement `Hash` instead of `splat`. However, this is not allowed since `splatting` in drivers is mostly used in 

```
queue(**({
      name: command,
    }.merge(options))) do
...
end
```

and queue's method is defined in `placeos-driver/src/placeos-driver.cr` as follows (line 114-117)
```
  # Queuing
  def queue(**opts, &block : Task -> Nil)
    @__queue__.add(**opts, &block)
  end
```
which disallows anything but splats and NamedTuple in the argument.

pkheav (changes requested):
I've had a look through and have picked up some things on style/feedback that @KimBurgess has given me previously. There are probably other things I have missed or not knowledgable enough to give feedback on so best to reach out to @KimBurgess for further feedback but this should hopefully be a good start of some things to look out for :)

PR #66, opened 2020-10-14
fix(bookings): pass empty Array instead of nil

pkheav:


caspiano (approved):

caspiano:
small note: it appears the email you have registered with your local git is not registered with github

pkheav:
Thanks @caspiano , did not notice that my email was not set and still not actually sure where it surfaces in github but I think I've set it up properly now.

PR #67, opened 2020-10-15
Fixes for place/bookings.cr for o365

pkheav:


pkheav:
These are the correct params https://github.com/PlaceOS/drivers/blob/bookings_fix/drivers/place/calendar.cr#L136-L143

pkheav:
Hey @stakach , I've noticed that the `owner` param isn't being used anywhere. https://github.com/PlaceOS/drivers/blob/4fb96ba75fb16513b54fe4bce77bfbd2ca1a2bcc/drivers/place/bookings.cr#L109

Is it supposed to be used for something or can I remove it?

stakach:
Yeah I think I was intending it for allowing another user to be the organizer of the booking
So we should probably do something like:

`host_calendar = owner.presence || @calendar_id`

If we want the calendar entry being owned by the specified account. Also in Google land the resource calendar cannot own a meeting, it needs to be assigned to the service account I think

pkheav:
Thanks @stakach , good to know about what google needs too.

For `book_now` 
https://github.com/PlaceOS/drivers/blob/8d8e401ea1ccce08f47d5cfc2036991437e886c4/drivers/place/bookings.cr#L109-L132

Do we want those meetings to be automatically started or should they just remain as pending? I can add these 2 lines in between lines 129-130 to do that but just wanted to check what functionality we want.
`@last_booking_started = event["event_start"]`
`define_setting(:last_booking_started, event["event_start"])`

pkheav:
@stakach I've made this commit
https://github.com/PlaceOS/drivers/pull/67/commits/d6ebd15fb0db50f4bbc87010576bf3e48e5f9bad

Logic is:
- For google setups, `setting?(String, :calendar_id)` should be defined as the service account which means `@calendar_id = service_account` and `room_is_organizer = false`
-  For o365 setups, `setting?(String, :calendar_id)` should not be defined which means `@calendar_id = system.email` and `room_is_organizer = true`

The room is always added as an attendee for both google/o365 setups.

pkheav:
@stakach , I'm calling `poll_events` here which works fine if the room is the owner of the booking.
https://github.com/PlaceOS/drivers/blob/bookings_fix/drivers/place/bookings.cr#L130

However, if we pass the `owner` param so another user can be the host/organizer of the booking, this creates the booking in the owner's calendar but takes a few seconds to actually also be in the room's calendar. This means the `poll_events` call on line 130 won't actually update `self[:bookings]` as the booking will not be in the room's calendar yet.

The 2 ways I can think of dealing with this are:
Option 1: Introduce a sleep period after creating the booking before calling `poll_events` on line 130 so that when `poll_events` is called, the booking does exist in the room's calendar.
Option 2: Manually insert the booking which we have from `event` here https://github.com/PlaceOS/drivers/blob/bookings_fix/drivers/place/bookings.cr#L119 into `self[:bookings]`

stakach (approved):
LGTM

PR #73, opened 2020-11-04
Hitachi projector

pkheav:


KimBurgess (commented):
Looks like this is still pretty fresh in the port. Would be good to tighten up some of the modelling and get the magic string and hashes of these into types.

pkheav:
Hey @KimBurgess , I'm not sure what else to do to cleanup the command forming as there doesn't seem to be a structure or consistent pattern with commands and they all seem pretty random to me. These are the docs if you want to have a look https://aca.im/driver_docs/Hitachi/cp-tw-series.pdf

KimBurgess:
The first few pages of that doc lay out the protocol structure. It looks like there are two levels to this:
1. Commands (which include a fixed header, CRC, action, type and code)
2. Which are they wrapped in a secondary layer for the IP comms. 

Where a clearly defined structure like this exists, drivers should model this rather than being a collection of preformed strings.

pkheav:
@KimBurgess I get that there is technically a structure according to the docs but there might as well not be as each command is pretty unique even for the same subset of commands like input selecting.

For example for input selecting, the CRC is different for HDMI1 and HDMI2 which is fine. However, their setting code is also different which doesn't really make such sense to me and why I think the structure in this protocol is useless IMO and also makes it harder to write nice structured code to build commands.

image

So to me, it seems like most commands need 4 bits to uniquely identify them which is 2 bits for CRC and 2 for setting code but annoyingly they are separated by 2 bits for Type which seems to be the same for all commands in a subset like input selecting. The only thing that makes sense in this protocol implementation to me is the 2 bits for Action which is [0x01, 0x00] for Set or [0x02, 0x00] for GET. This is why I'm not sure it's worth cleaning up the command building in the driver because of this.

pkheav:
Writing this all out, I can see what you're saying though @KimBurgess so I'll give it a try. Type can be used to identify what subset of commands (e.g. input switching) is being used and CRC+Setting code can be used to represent the specific command of a subset (e.g. HDMI)

pkheav:
I've made this commit to show my approach on this. Let me know if you think this better before I fully commit to refactoring this @KimBurgess https://github.com/PlaceOS/drivers/commit/cd1621b4975d038d1b7380612dc43d1989a5ce9e

KimBurgess:
@pkheav - sorry for the delayed response - just catching up on some of the PR's.

You can likely break that down a little more. If you look at the protocol structure there are three components to each command:

1. `Action`
This is one of: `Set`, `Get`, `Increment`, `Decrement`, or `Execute` and can pipe directly into an enum.
```crystal
enum Action
  Set       = 0x0100
  Get       = 0x0200
  Increment = 0x0400
  Decrement = 0x0500
  Execute   = 0x0600
end
```

2. `Type`

3. `Setting Code`

KimBurgess:
Unintended keyboard mash... reopening.

drivers/hitachi/projector/cp_tw_series_basic.cr
```diff
@@ -0,0 +1,230 @@
...
+    picture_mute: "CD F0 02 00 A0 20 00 00",
+    lamp:         "C2 FF 02 00 90 10 00 00",
+    filter:       "C2 F0 02 00 A0 10 00 00",
+  }
+  {% for name, data in QueryRequests %}
+    @[Security(Level::Administrator)]
+    def {{name.id}}?(**options)
+      do_send({{data}}, **options, name: {{name.id.stringify}} + '?')
+    end
+  {% end %}
```
caspiano, line 123: It might make it a bit clearer and less prone to error to have all of the messages collected in a single place.
the do send method [in this driver](https://github.com/PlaceOS/drivers/blob/8c505742bbac53180d3ecdbff842e160277fbebf/drivers/panasonic/projector/nt_control.cr#L247) is pretty neat, and given that it's a  NamedTuple, invalid keys to the hash are checked at compile time.


Then these lines could look something like this

```suggestion
  QueryRequests = %i(power input error freeze audio_mute picture_mute lamp filter)
  {% for name in QueryRequests %}
    @[Security(Level::Administrator)]
    def {{name.id}}?(**options)
      do_send({{ name }}, **options, name: {{name.id.stringify}} + '?')
    end
  {% end %}
```

caspiano (commented):
clear work so far!

drivers/hitachi/projector/cp_tw_series_basic.cr
caspiano: kim mentioned the enum manner of doing things as well, which is good too

drivers/hitachi/projector/cp_tw_series_basic_spec.cr
```diff
@@ -0,0 +1,86 @@
...
+  q = {
+    power:        "19 D3 02 00 00 60 00 00",
+    input:        "CD D2 02 00 00 20 00 00",
+    error:        "D9 D8 02 00 20 60 00 00",
+    freeze:       "B0 D2 02 00 02 30 00 00",
+    audio_mute:   "75 D3 02 00 02 20 00 00",
+    picture_mute: "CD F0 02 00 A0 20 00 00",
+    lamp:         "C2 FF 02 00 90 10 00 00",
+    filter:       "C2 F0 02 00 A0 10 00 00",
+  }
```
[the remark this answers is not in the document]
pkheav: Hmm, yeah I was trying to figure out how to access `QueryRequests` from the driver but I've honestly got no idea.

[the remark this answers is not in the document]
pkheav: okay never mind I figured this out https://github.com/PlaceOS/drivers/pull/73/commits/3daedc76221857ae8ead9cfb310ad98063440f2b

drivers/hitachi/projector/cp_tw_series_basic.cr
pkheav: addressed with https://github.com/PlaceOS/drivers/pull/73/commits/07bb5c0232a9c1012de0b76be39a07ff8c26b803

pkheav:
> you probably could skip the `def self.query_requests` by doing
> 
> ```
>    q = Hitachi::Projector::CpTwSeriesBasic::QueryRequests 
> ```

Nice, thanks @caspiano , did not know that was possible and way neater :)
https://github.com/PlaceOS/drivers/pull/73/commits/05cc8fd1190b640bb15bab0ebe168f094f1c5437

caspiano (approved):
lgtm

PR #74, opened 2020-11-04
feat(mdc): set targets to nil once stable

pkheav:
I think this PR/commit should help address this issue @KimBurgess https://github.com/PlaceOS/drivers/pull/19#pullrequestreview-508000238


KimBurgess (commented):
There's a double up here between the `*_stable` and `*_target` vars now that these are nilable. This can likely be simplified so that if a target is `nil`, it is stable.

PR #76, opened 2020-11-06
Global cache

pkheav:
Just going to open up an initial PR to get some feedback from you @KimBurgess when you can please. Not 100% sure what this device actually does but I think it's like a router/hub that you connect IO devices too I think?


KimBurgess:
@pkheav unfortunately I haven't had time to review this - hopefully another member of the @PlaceOS/drivers team may have bandwidth soon, but to answer you're query: yep, you are correct in that assumption re functionality. These are a device that provide an IP -> RS-232 gateway and IP controllable relays commonly used for providing connectivity with devices that do not have in-built network connectivity for control.

https://www.globalcache.com/products/gc-100/

jeremyw24:
@stakach can you please review this PR and merge if all good?

PR #79, opened 2020-11-10
fix(place/area/config): allow zone.metadata.area.properties.Hash

w-le:
support

"areas": [
        {
            ...
            "properties": {
                "label_location": {
                    "x": 0.5089,
                    "y": 0.6673
                }
            }


stakach (approved):
LGTM

PR #82, opened 2020-11-18
fix(mdc): correct typo

pkheav:
I think this is a typo as I don't think `!!` means anything in crystal


pkheav:
Closing this PR as it's also failing to compile and seems like it's intended for `power?` to return a `Bool` instead of `Bool?`

PR #88, opened 2020-11-30
fix(mdc): calculate correct checksums

### LINKED ISSUE #87: Bug: Samsung MCD driver checksum calculation is incorrect
**Describe the bug**
The way the checksum is being calculated in the Samsung MDC driver is incorrect as are the return values in the tests.

**To Reproduce**
The correct response to a power query (where the power is ON) is `responds("\xAA\xFF#{id}\x03\x41\xF9\x00\x3c")`

When running the test with the correct return value the module will log an error `"Invalid checksum, checksum should be: ff"`

pkheav:
fixes #87


stakach (approved):
LGTM

pkheav:
Nice, thanks @stakach

PR #89, opened 2020-12-01
fix: Extron matrix driver

KimBurgess:
Resolves type issues and a missing response parser within the Extron Matrix driver.

The branch state in the originally merged PR was missing these. Due to the current lack of automated testing during merges, this would not have been caught.

### DIFF
--- a/drivers/extron/matrix.cr
+++ b/drivers/extron/matrix.cr
@@ -81,7 +81,7 @@ class Extron::Matrix < PlaceOS::Driver
   end
 
   private def send(command, parser : SIS::Response::Parser(T)) forall T
-    send command, &.itself
+    send command, parser, &.itself
   end
 
   # Response callback for async responses.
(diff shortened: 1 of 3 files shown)


pkheav:
@KimBurgess Not sure why I'm getting this error when checking out that branch. I've run `shards install`, `docker-compose down` and `docker-compose up -d` in case that's needed to get that parser dependency 
image

KimBurgess:
Your local `placeos/drivers` docker image will be missing that shard.

From within the branch:
```bash
docker build -t placeos/drivers .
```

Then reload it with
```bash
docker-compose up -d
```

pkheav:
Oh right, I have to worry about the docker image too. Thanks for the tip @KimBurgess , also nice that a nicer/newer UI for the driver test runner comes with it. I've got that running now and all tests are passing so looks good to me so I'll approve.

pkheav (approved):

PR #106, opened 2021-02-10
feat: Kramer switches PPT-367

### LINKED ISSUE #47: Kramer Switchers

pkheav:
I've ported 2/3 [drivers from here](https://github.com/acaprojects/ruby-engine-drivers/tree/beta/modules/kramer/switcher) but not [here](https://github.com/acaprojects/ruby-engine-drivers/blob/beta/modules/kramer/switcher/presentation.rb) as it is just a simple extension of the driver p3000 driver. 

I can port the presentation.rb driver if needed though as it is simple.

Closes #47


stakach (changes requested):
These drivers should implement the [Switchable](https://github.com/PlaceOS/driver/blob/master/src/placeos-driver/interface/switchable.cr#L9) interface

@kimburgess we should probably update your [Extron Matrix driver](https://github.com/PlaceOS/drivers/blob/master/drivers/extron/matrix.cr) to do the same

The switchable interface was brainstormed early on, so happy for updates to made to the interface too.

jeremyw24:
@stakach can you please review this?

PR #112, opened 2021-03-02
feat: add Wolfvision eye14 drivers PPT-368

polonski:
- [x]  Wolfvision eye14 drivers rebuilt in Crystal.


drivers/wolfvision/eye14.cr
```diff
@@ -0,0 +1,215 @@
...
+  @zoom_range = 0..3923
+  @iris_range = 0..4094
+
+  tcp_port 50915 # Need to go through an RS232 gatway
+  descriptive_name "WolfVision EYE-14"
+  generic_name :Camera
+
+  COMMANDS = {
+    power_on:        "\\x01\\x30\\x01\\x01",
+    power_off:       "\\x01\\x30\\x01\\x00",
```
stakach, line 32: these commands should be binary, i.e. "\x01\x30\x01\x00"
however that might cause issues in crystal as all strings are UTF8 whereas the ruby code is ASCII (no escape characters) - Ruby also uses binary strings for most IO, whereas crystal has `Bytes` which might be more appropriate in this case

stakach (changes requested):
looks good
please take a look at the requested change and we can merge

drivers/wolfvision/eye14.cr
polonski: binary commands implemented as suggested. Conversion to `Bytes` is occurring as expected.

stakach (changes requested):
I think using the Ruby Driver here as the basis of this driver was a mistake, from the look of it the Ruby driver probably never worked and was based on guess work related to the linked protocol description.

After a quick google search I found Wolfvision has since released a much more detailed protocol description:
https://www.wolfvision.com/wolf/protocol_command_wolfvision/protocol_command.htm#t=instructions%2Fpacket_structure.htm

This indicates that we should probably be starting from scratch with this driver based on that protocol breakdown.

I would also recommend using https://github.com/spider-gazelle/bindata
for describing this binary structure

something like

```crystal
class WolfPacket < BinData
  endian big
  
  bit_field do
    bool :error, default: false
    bits 3, :head_reserved
    bool :two_byte_cmd, default: false
    bool :two_byte_length, default: false
    bool :two_byte_header, default: false
    bool :set_cmd, default: false
  end

  bit_field onlyif: ->{ two_byte_header } do
    bits 7, :ext_head_reserved
    bool :four_byte_length, default: false
  end

  uint8 :short_cmd, onlyif: ->{ !two_byte_cmd }
  uint16 :long_cmd, onlyif: ->{ two_byte_cmd }

  def command : UInt16
    two_byte_cmd ? long_cmd : short_cmd.to_u16
  end

  def command=(number : Int)
    if number > UInt8::MAX
      self.long_cmd = number.to_u16
      self.two_byte_cmd = true
    else
      self.short_cmd = number.to_u8
      self.two_byte_cmd = false
    end
    number
  end

  uint8 :byte_len, onlyif: ->{ !four_byte_length && !two_byte_length }
  uint16 :short_len, onlyif: ->{ !four_byte_length && two_byte_length }
  uint32 :long_len, onlyif: ->{ four_byte_length }

  def length : UInt32
    four_byte_length ? long_len : (two_byte_length ? short_len.to_u32 : byte_len.to_u32)
  end

  def length=(number : Int)
    if number > UInt16::MAX
      self.four_byte_length = number.to_u32
      self.two_byte_header = true
      self.two_byte_length = false
      self.four_byte_length = true
    elsif number > UInt8::MAX
      self.short_len = number.to_u16
      self.two_byte_header = false
      self.two_byte_length = true
      self.four_byte_length = false
    else
      self.byte_len = number.to_u8
      self.two_byte_header = false
      self.two_byte_length = false
      self.four_byte_length = false
    end
    number
  end

  bytes :payload, length: ->{ length }, onlyif: ->{ !error }
end
```

PR #118, opened 2021-03-15
feat(place): event scrape

pkheav:
Looks good to me testing on my local partner environment. Let me know if anything else should be changed or added @w-le @stakach 

image

Also @stakach , might need your help figuring out why these methods https://github.com/PlaceOS/drivers/blob/0e234bc95e7c0af2df76c33458f9b91886b9745f/drivers/place/staff_api.cr#L225-L242 don't work as they would be cleaner and probably be a useful addition to `staff_api.cr` as they're more general than the working methods I added instead https://github.com/PlaceOS/drivers/blob/0e234bc95e7c0af2df76c33458f9b91886b9745f/drivers/place/staff_api.cr#L189-L219.

Oh yeah, tried writing a spec but couldn't get it working as it seems like the spec runner is a little bit flaky.


drivers/place/event_scrape.cr
```diff
@@ -0,0 +1,84 @@
...
+      end
+    end
+
+    response
+  end
+
+  def get_system_bookings(sys_id : String, start_epoch : Int64?, end_epoch : Int64?) : Array(Event)
+    booking_module = staff_api.modules_from_system(sys_id).get.as_a.find { |mod| mod["name"] == "Bookings" }
+    # If the system has a booking module with bookings
+    if booking_module && (bookings = staff_api.get_module_state(booking_module["id"].as_s).get["bookings"]?)
```
kimburgess, line 65: You may need to expand the staff API driver to support this, but this can collapse into a single query via this endpoint:

https://github.com/PlaceOS/rest-api/blob/5d3acc0cff8d949d360a8f6de9c2e916c52f5cf1/src/placeos-rest-api/controllers/systems.cr#L314-L321

kimburgess (changes requested):

PR #119, opened 2021-03-16
Freespace driver

stakach:
init implementation, requires real world testing but should be good to merge


drivers/freespace/sensor_api.cr
```diff
@@ -0,0 +1,243 @@
...
+    )
+    logger.debug { "login response: #{response.body}" }
+    raise "issue obtaining token: #{response.status_code}\n#{response.body}" unless response.success?
+
+    # auth key is valid for 5 seconds
+    schedule.in(5.seconds) { @auth_key = nil }
+    @auth_key = response.headers["X-Auth-Key"]
+  end
+
+  def received(bytes, task)
```
caspiano, line 162: Should this be public?

drivers/freespace/sensor_api.cr
```diff
@@ -0,0 +1,243 @@
...
+        {
+          location:    activity.capacity == 1 ? "desk" : "area",
+          at_location: activity.state,
+          map_id:      activity.name,
+          level:       loc_details[:level_id],
+          building:    loc_details[:building_id],
+          capacity:    activity.capacity,
+        }
+      end
+    }.flatten
```
caspiano, line 241:
```suggestion
    loctions.flat_map do |loc_id|
      location_id = loc_id.to_i64
      loc_details = @floor_mappings[loc_id]

      @space_state.values.compact_map do |activity|
        next if activity.location_id != location_id || activity.state == 0 || activity.capacity > 1

        {
          location:    activity.capacity == 1 ? "desk" : "area",
          at_location: activity.state,
          map_id:      activity.name,
          level:       loc_details[:level_id],
          building:    loc_details[:building_id],
          capacity:    activity.capacity,
        }
      end
    end
```

drivers/freespace/sensor_api.cr
stakach: it's a protected method so not exposed to the public API and only used internally by `DriverManager`

caspiano (approved):

PR #122, opened 2021-03-22
Cisco ise

pkheav:


drivers/cisco/ise/guests.cr
```diff
@@ -0,0 +1,152 @@
...
+    phone_number : String = "0123456789", # Mandatory, use a fake value as default
+    sms_service_provider : String? = nil, # Use this param to override the setting
+    guest_type : String? = nil,           # Mandatory but use this param to override the setting
+    portal_id : String? = nil             # Mandatory but use this param to override the setting
+  )
+    # Determine the name of the attendee for ISE
+    guest_names = attendee_name.split
+    first_name_index_end = guest_names.size > 1 ? -2 : -1
+    first_name = guest_names[0..first_name_index_end].join(' ')
+    last_name = guest_names[-1]
```
w-le, line 65: took me a minute to grok this but yeah that will work.
FYI I think when the input name is a string with no spaces then last_name would be the same as first_name instead of being null - that may be fine for your use case as long as you're aware that it could happen

w-le (approved):
lgtm

drivers/cisco/ise/guests.cr
pkheav: Sorry it also confused me a bit looking over at it again especially because of the negative indexes with first_name_index_end but I think it is the cleanest way to do this. I probably should of added a comment.

pkheav: I think I accepted that last_name would be the same as first_name in a case where the `attendee_name = "Phil"` since that would be pretty rare I think

PR #128, opened 2021-03-31
Sony LCD

### LINKED ISSUE #104: Sony LCD > Migrate Ruby Driver

tassja:
Closes #104

### DIFF
--- a/drivers/sony/displays/bravia.cr
+++ b/drivers/sony/displays/bravia.cr
@@ -66,13 +66,13 @@ class Sony::Displays::Bravia < PlaceOS::Driver
 
   def switch_to(input : Inputs)
     logger.debug { "switching input to #{input}" }
-    request(:input, input.to_message)
+    request(Command::Input, input.to_message)
     self[:input] = input.to_s
     input?
   end
 
   def input?
-    query(:input, priority: 0)
+    query(Command::Input, priority: 0)
   end
 
   def on_load
@@ -91,13 +91,13 @@ class Sony::Displays::Bravia < PlaceOS::Driver
   end
 
   def power(state : Bool)
-    request(:power, state)
+    request(Command::Power, state)
     logger.debug { "Sony display requested power #{state ? "on" : "off"}" }
     power?
   end
 
   def power?
-    query(:power)
+    query(Command::Power)
   end
 
   def mute(
@@ -105,7 +105,7 @@ class Sony::Displays::Bravia < PlaceOS::Driver
     index : Int32 | String = 0,
     layer : MuteLayer = MuteLayer::AudioVideo
   )
-    request(:mute, state)
+    request(Command::Mute, state)
     mute?
   end
 
@@ -114,11 +114,11 @@ class Sony::Displays::Bravia < PlaceOS::Driver
   end
 
   def mute?
-    query(:mute, priority: 0)
+    query(Command::Mute, priority: 0)
   end
 
   def mute_audio(state : Bool = true)
-    request(:audio_mute, state)
+    request(Command::AudioMute, state)
     audio_mute?
   end
 
@@ -127,16 +127,16 @@ class Sony::Displays::Bravia < PlaceOS::Driver
   end
 
   def audio_mute?
-    query(:audio_mute, priority: 0)
+    query(Command::AudioMute, priority: 0)
   end
 
   def volume(level : Int32)
-    request(:volume, level.to_i)
+    request(Command::Volume, level.to_i)
     volume?
   end
 
   def volume?
-    query(:volume, priority: 0)
+    query(Command::Volume, priority: 0)
   end
 
   def do_poll
@@ -236,14 +236,14 @@ end
   end
 
   protected def request(command, parameter, **options)
-    cmd = COMMANDS[command]
+    cmd = command.function
     parameter = parameter ? 1 : 0 if parameter.is_a?(Bool)
     param = parameter.to_s.rjust(16, '0')
     do_send(MessageType::Control, cmd, param, **options)
   end
 
   protected def query(state, **options)
-    cmd = COMMANDS[state]
+    cmd = state.function
     do_send(MessageType::Enquiry, cmd, HASH, **options)
   end
 
--- a/drivers/sony/displays/bravia.cr
+++ b/drivers/sony/displays/bravia.cr
@@ -9,20 +9,70 @@ class Sony::Displays::Bravia < PlaceOS::Driver
   INDICATOR = "\x2A\x53" # *S
   msg_length = 21
 
+  # INPUTS = {
+  #   "Tv"     => "00000",
+  #   "Hdmi"   => "10000",
+  #   "Mirror" => "50000",
+  #   "Vga"    => "60000",
+  # }
+  # INPUT_LOOKUP = INPUTS.invert
+
+  enum Inputs
+    Tv
+    Tv1
+    Tv2
+    Tv3
+    Hdmi
+    Hdmi1
+    Hdmi2
+    Hdmi3
+    Mirror
+    Mirror1
+    Mirror2
+    Mirror3
+    Vga
+    Vga1
+    Vga2
+    Vga3
+  end
+
   INPUTS = {
-    "Tv"     => "00000",
-    "Hdmi"   => "10000",
-    "Mirror" => "50000",
-    "Vga"    => "60000",
+    Inputs::Tv      => "Tv",
+    Inputs::Tv1     => "Tv1",
+    Inputs::Tv2     => "Tv2",
+    Inputs::Tv3     => "Tv3",
+    Inputs::Hdmi    => "Hdmi",
+    Inputs::Hdmi1   => "Hdmi1",
+    Inputs::Hdmi2   => "Hdmi2",
+    Inputs::Hdmi3   => "Hdmi3",
+    Inputs::Mirror  => "Mirror",
+    Inputs::Mirror1 => "Mirror1",
+    Inputs::Mirror2 => "Mirror2",
+    Inputs::Mirror3 => "Mirror3",
+    Inputs::Vga     => "Vga",
+    Inputs::Vga1    => "Vga1",
+    Inputs::Vga2    => "Vga2",
+    Inputs::Vga3    => "Vga3",
   }
   INPUT_LOOKUP = INPUTS.invert
 
-  def switch_to(input : String)
+  INPUTS_CONVERT = {
+    "Tv"     => "0000",
+    "Hdmi"   => "1000",
+    "Mirror" => "5000",
+    "Vga"    => "6000",
+  }
+
+  CONVERT_LOOKUP = INPUTS_CONVERT.invert
+
+  include Interface::InputSelection(Inputs)
+
+  def switch_to(input : Inputs)
     parsed_input = input.to_s.scan(/[^0-9]+|\d+/)
     index = parsed_input.size < 2 ? "1" : parsed_input[1][0]
-    input_type = INPUTS[parsed_input[0][0].capitalize]
-    request(:input, "#{input_type}#{index.rjust(4, '0')}")
-    self[:input] = "#{input_type}#{index}" # for a responsive UI
+    input_joined = INPUTS_CONVERT[parsed_input[0][0]] + index.rjust(5, '0')
+    request(:input, input_joined)
+    self[:input] = "#{input}"
     input?
   end
 
@@ -201,12 +251,12 @@ class Sony::Displays::Bravia < PlaceOS::Driver
     when :mac_address
       self[:mac_address] = parsed_data.split('#')[0]
     when :input
-      input_num = convert_binary(param[7..11])
-      index_num = convert_binary(param[12..-1]).to_i
+      input_num = convert_binary(param[7..10])
+      index_num = convert_binary(param[11..-1]).to_i
       if index_num == 1
-        self[:input] = INPUT_LOOKUP[input_num]
+        self[:input] = CONVERT_LOOKUP[input_num]
       else
-        self[:input] = "#{INPUT_LOOKUP[input_num]}#{index_num}"
+        self[:input] = "#{CONVERT_LOOKUP[input_num]}#{index_num}"
       end
     end
   end
--- a/drivers/sony/displays/bravia.cr
+++ b/drivers/sony/displays/bravia.cr
@@ -187,7 +187,7 @@ class Sony::Displays::Bravia < PlaceOS::Driver
   TYPE_RESPONSE = TYPES.to_h.invert
 
   protected def convert_binary(data)
-    data.map { |x| x.chr }.join
+    data.join &.chr
   end
 
   protected def request(command, parameter, **options)
--- a/drivers/sony/displays/bravia.cr
+++ b/drivers/sony/displays/bravia.cr
@@ -198,38 +198,38 @@ class Sony::Displays::Bravia < PlaceOS::Driver
     broadcast_address: "BADR",
     mac_address:       "MADR",
   }
+
   {% begin %}
-
-enum Command
-  {% begin %}
-    {% for command in COMMANDS.keys %}
-      {{ command.camelcase.id }}
-    {% end %}
-  {% end %}
-
-  def function
+  enum Command
     {% begin %}
-    case self
-    {% for kv in COMMANDS.to_a %}
-    	{% command, value = kv[0], kv[1] %}
-      in {{ command.camelcase }} then {{ value }}
+      {% for command in COMMANDS.keys %}
+        {{ command.camelcase.id }}
+      {% end %}
     {% end %}
-  	end
-		{% end %}
-  end
 
-  def self.from_response?(message)
-    {% begin %}
-    case message
-    {% for kv in COMMANDS.to_a %}
-    {% command, value = kv[0], kv[1] %}
-    when {{ value }} then {{ command.camelcase.id }}
-    {% end %}
+    def function
+      {% begin %}
+      case self
+      {% for kv in COMMANDS.to_a %}
+        {% command, value = kv[0], kv[1] %}
+          in {{ command.camelcase }} then {{ value }}
+      {% end %}
+      end
+      {% end %}
     end
-    {% end %}
+
+    def self.from_response?(message)
+      {% begin %}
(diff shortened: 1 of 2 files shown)


drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,213 @@
...
+require "placeos-driver/interface/switchable"
+
+class Sony::Displays::Bravia < PlaceOS::Driver
+  include Interface::Powerable
+  include Interface::Muteable
+
+  INDICATOR = "\x2A\x53" # *S
+  msg_length = 21
+
+  INPUTS = {
```
stakach, line 12: We would want to use the Switchable interface inputs, this is how it's done for a [Samsung driver](https://github.com/PlaceOS/drivers/blob/master/drivers/samsung/displays/mdc_protocol.cr#L11-L31)

stakach (changes requested):
Other than the switchable interface missing, everything looks great :)

caspiano:
Looking good so far!

drivers/sony/displays/bravia.cr
```diff
@@ -9,20 +9,70 @@ class Sony::Displays::Bravia < PlaceOS::Driver
...
+    Inputs::Mirror  => "Mirror",
+    Inputs::Mirror1 => "Mirror1",
+    Inputs::Mirror2 => "Mirror2",
+    Inputs::Mirror3 => "Mirror3",
+    Inputs::Vga     => "Vga",
+    Inputs::Vga1    => "Vga1",
+    Inputs::Vga2    => "Vga2",
+    Inputs::Vga3    => "Vga3",
   }
   INPUT_LOOKUP = INPUTS.invert
```
stakach, line 51: doesn't look like you need or use `INPUT_LOOKUP` or `INPUTS` anymore

stakach (changes requested):
all good, I think `INPUT_LOOKUP` and `INPUTS` should be removed

stakach (dismissed):
LGTM

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,230 @@
...
+    65 => "A",
+    69 => "E",
+    67 => "C",
+    78 => "N",
+  }
+
+  TYPE_RESPONSE = TYPES.to_h.invert
+
+  protected def convert_binary(data)
+    data.map { |x| x.chr }.join
```
caspiano, line 190:
```suggestion
    data.join &.chr
```

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,230 @@
...
+require "placeos-driver/interface/switchable"
+
+class Sony::Displays::Bravia < PlaceOS::Driver
+  include Interface::Powerable
+  include Interface::Muteable
+
+  INDICATOR = "\x2A\x53" # *S
+  HASH      = "################"
+  ERROR     = 70
+  msg_length = 21
```
caspiano, line 12: Where's this being used? also if this is a constant, it should be uppercase

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,229 @@
...
+    "Mirror" => "5000",
+    "Vga"    => "6000",
+  }
+
+  CONVERT_LOOKUP = INPUTS_CONVERT.invert
+
+  include Interface::InputSelection(Inputs)
+
+  def switch_to(input : Inputs)
+    parsed_input = input.to_s.scan(/[^0-9]+|\d+/)
```
caspiano, line 44: Is this extracting the index? If so, this could possibly be done in a simpler non-regex based manner

drivers/sony/displays/bravia.cr
```diff
@@ -193,16 +183,16 @@ class Sony::Displays::Bravia < PlaceOS::Driver
...
 
   protected def query(state, **options)
     cmd = COMMANDS[state]
-    do_send(:enquiry, cmd, HASH, **options)
+    do_send(1, cmd, HASH, **options)
   end
 
   protected def do_send(type, command, parameter, **options)
-    cmd_type = TYPES[type]
+    cmd_type = type == 0 ? MessageType::Control.to_i64.chr : MessageType::Enquiry.to_i64.chr
```
caspiano: Maybe make this a method on the enum `#control_char`

drivers/sony/displays/bravia.cr
```diff
@@ -132,18 +132,17 @@ class Sony::Displays::Bravia < PlaceOS::Driver
...
-
-    case TYPE_RESPONSE[type]
-    when :answer
+    type = MessageType.from_value(data[2]).to_i64.chr
+    case type
+    when 'A'
       update_status cmd, param
       task.try &.success
-    when :notify
+    when 'N'
```
caspiano:
```suggestion
    case MessageType.from_value(data[2])
    when .answer?
      update_status cmd, param
      task.try &.success
    when .notify?
```

drivers/sony/displays/bravia.cr
```diff
@@ -169,21 +168,12 @@ class Sony::Displays::Bravia < PlaceOS::Driver
...
-    67 => "C",
-    78 => "N",
-  }
-
-  TYPE_RESPONSE = TYPES.to_h.invert
+  enum MessageType : UInt8
+    Answer  = 0x41
+    Control = 0x43
+    Enquiry = 0x45
+    Notify  = 0x4e
```
caspiano, line 175:
```suggestion
    Notify  = 0x4e
    
    def control_character
      case self
      in Answer  then "A"
      in Control then "C"
      in Enquiry then "E"
      in Notify  then "N"
      end
    end
```

drivers/sony/displays/bravia.cr
```diff
@@ -193,16 +183,16 @@ class Sony::Displays::Bravia < PlaceOS::Driver
...
+    do_send(0, cmd, param, **options)
   end
 
   protected def query(state, **options)
     cmd = COMMANDS[state]
-    do_send(:enquiry, cmd, HASH, **options)
+    do_send(1, cmd, HASH, **options)
   end
 
   protected def do_send(type, command, parameter, **options)
```
caspiano, line 194: Could you please add some types to this interface?

drivers/sony/displays/bravia.cr
```diff
@@ -137,12 +137,12 @@ class Sony::Displays::Bravia < PlaceOS::Driver
     param = data[7..-1]
 
     return task.try(&.abort("error")) if param.first? == ERROR
-    type = MessageType.from_value(data[2]).to_i64.chr
-    case type
-    when 'A'
+
+    case MessageType.from_value(data[2])
```
caspiano, line 141: Actually, `from_value?` is better here. Then logging `data[2]` in the else clause below would be a good idea if we want to diagnose malformed messages

drivers/sony/displays/bravia.cr
kimburgess: Those values are the characters. If you want them directly [`Int#chr`](https://crystal-lang.org/api/1.0.0/Int.html#chr-instance-method) is available. This works well if you are serialising elsewhere (e.g. when building a command packet before transmission).
```crystal
MessageType::Answer.value.chr # => 'A'
```

If the primary serialisation is the protocol form (e.g. these are not appear in a bunch of log messages or similar) a nice pattern for this though to pop a `#to_s(IO)` override straight on the enum:
```crystal
enum MessageType : UInt8
  Answer  = 0x41
  Control = 0x43
  Enquiry = 0x45
  Notify  = 0x4e

  def to_s(io)
    io << self.value.chr
  end
end

String.build { |io| io << MessageType::Answer }  # => "A"

"Foo#{MessageType::Answer}" # => "FooA"

File: drivers/sony/displays/bravia.cr [hunk 11]
```diff
@@ -0,0 +1,226 @@
...
+    case cmd
+    when :power, :mute, :audio_mute, :pip
+      self[cmd] = parsed_data.to_i == 1
+    when :volume
+      self[:volume] = parsed_data.to_i
+    when :mac_address
+      self[:mac_address] = parsed_data.split('#')[0]
+    when :input
+      input_num = convert_binary(param[7..10])
+      index_num = convert_binary(param[11..-1]).to_i
```
caspiano, line 218: You already have done the conversion up on 208, index into `parsed_data` here.

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,226 @@
...
+      if index_num == 1
+        self[:input] = CONVERT_LOOKUP[input_num]
```
caspiano, line 220: Does the `input_num` vary here for members of the input types?
Since this is related to `Input`, it would be neater if the logic was held inside the `Input` enum.

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,226 @@
...
+    tv_input:          "ISRC",
+    input:             "INPT",
+    toggle_mute:       "TPMU",
+    pip:               "PIPI",
+    toggle_pip:        "TPIP",
+    position_pip:      "TPPP",
+    broadcast_address: "BADR",
+    mac_address:       "MADR",
+  }
+  RESPONSES = COMMANDS.to_h.invert
```
caspiano, line 184: Could be cleaner if contained in an enum. That would allow for total matching on the `case` statement down on 209-223.
Pragmatically, however, I think this is in a good place.

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,226 @@
...
+    Vga2
+    Vga3
+  end
+
+  INPUTS_CONVERT = {
+    "Tv"     => "0000",
+    "Hdmi"   => "1000",
+    "Mirror" => "5000",
+    "Vga"    => "6000",
+  }
```
caspiano, line 42: I'd move this into `Input`

drivers/sony/displays/bravia.cr
caspiano: This would be good to address in a cleaner way, as well as isolated within `Input`

caspiano (commented):
A few small things to address and we're good to go

drivers/sony/displays/bravia.cr
```diff
@@ -207,20 +224,21 @@ class Sony::Displays::Bravia < PlaceOS::Driver
...
       self[:mac_address] = parsed_data.split('#')[0]
     when :input
-      input_num = parsed_data[7..10]
-      index_num = parsed_data[11..-1].to_i
-      if index_num == 1
-        self[:input] = CONVERT_LOOKUP[input_num]
-      else
-        self[:input] = "#{CONVERT_LOOKUP[input_num]}#{index_num}"
-      end
+      self[:input] = Inputs.from_message?(parsed_data[7], parsed_data[15].to_i)
```
caspiano, line 241: No need to `to_i` here

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,274 @@
...
+  def function
+    {% begin %}
+    case self
+    {% for kv in COMMANDS.to_a %}
+    	{% command, value = kv[0], kv[1] %}
+      in {{ command.camelcase }} then {{ value }}
+    {% end %}
+  	end
+		{% end %}
+  end
```
caspiano, line 219: Pls convert this file to use spaces

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,274 @@
...
+    Enquiry = 0x45
+    Notify  = 0x4e
+
+    def control_character
+      case self
+      in Answer  then "A"
+      in Control then "C"
+      in Enquiry then "E"
+      in Notify  then "N"
+      end
```
caspiano, line 163:
```suggestion
      value.chr
```

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,274 @@
...
+    end
+  end
+
+  def received(data, task)
+    parsed_data = convert_binary(data[3..6])
+    cmd = Command.from_response?(parsed_data)
+    param = data[7..-1]
+    return task.try(&.abort("error")) if cmd.nil?
```
caspiano, line 171:
```suggestion
    return task.try(&.abort("unrecognised command: #{parsed_data}") if cmd.nil?
    param = data[7..-1]
```

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,274 @@
...
+    return task.try(&.abort("error")) if param.first? == ERROR
+    case MessageType.from_value?(data[2])
+    when MessageType::Answer
+      update_status cmd, param
+      task.try &.success
+    when MessageType::Notify
+      update_status cmd, param
+    else
+      logger.debug { "Unhandled device response: #{data[2]}" }
```
caspiano, line 180:
```suggestion
      logger.debug { "Unhandled device response: #{data[2].chr rescue data[2]}" }
```

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,269 @@
...
+        {% end %}
+      end
+      {% end %}
+    end
+  end
+  {% end %}
+
+  protected def convert_binary(data)
+    data.join &.chr
+  end
```
caspiano, line 231: This could be replaced with `String.new(bytes: data)`

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,269 @@
...
+      case self
+      when Hdmi1, Hdmi2, Hdmi3
+        "1000"
+      when Mirror1, Mirror2, Mirror3
+        "5000"
+      when Vga1, Vga2, Vga3
+        "6000"
+      else
+        "0000"
+      end
```
caspiano, line 46: When we have the potential for exhaustive case via `in`, it's preferred over `when`

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,269 @@
+require "placeos-driver/interface/powerable"
+require "placeos-driver/interface/muteable"
+require "placeos-driver/interface/switchable"
+
+class Sony::Displays::Bravia < PlaceOS::Driver
+  include Interface::Powerable
+  include Interface::Muteable
+
+  INDICATOR = "\x2A\x53" # *S
+  HASH      = "################"
```
kimburgess, line 10: These can both move into `#do_send` inline as part of the serialisation.

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,269 @@
...
+             when '6' then "Vga"
+             end
+
+      "#{type}#{index}" if type
+    end
+
+    def to_message : String
+      "#{type_hint}#{(index).to_s.rjust(5, '0')}"
+    end
+  end
```
kimburgess, line 63:
```suggestion
  enum Input : UInt32
    {% for idx in 1..3 %}
      Tv{{idx}}     = {{             idx}}
      Hdmi{{idx}}   = {{10000_0000 + idx}}
      Mirror{{idx}} = {{50000_0000 + idx}}
      Vga{{idx}}    = {{60000_0000 + idx}}
    {% end %}

    def self.from_param(value : String) : self
      from_value UInt32.new(value)
    rescue
      raise "Unknown enum #{self} value: #{value}"
    end

    def to_param : String
      value.to_s.rjust(15, '0')
    end
  end
```

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,269 @@
...
+
+  include Interface::InputSelection(Inputs)
+
+  def switch_to(input : Inputs)
+    logger.debug { "switching input to #{input}" }
+    request(Command::Input, input.to_message)
+    self[:input] = input.to_s
+    input?
```
kimburgess, line 71: A lot of these control methods looks to be running a second command. This should be needed. Responses from control commands will enable you to update state if there was a success response.

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,269 @@
...
+    end
+  end
+
+  def disconnected
+    schedule.clear
+  end
+
+  def power(state : Bool)
+    request(Command::Power, state)
+    logger.debug { "Sony display requested power #{state ? "on" : "off"}" }
```
kimburgess, line 95:
```suggestion
```

drivers/sony/displays/bravia.cr
caspiano: i think these are fine, just make them private constants so they don't show up in documentation

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,269 @@
...
+require "placeos-driver/interface/muteable"
+require "placeos-driver/interface/switchable"
+
+class Sony::Displays::Bravia < PlaceOS::Driver
+  include Interface::Powerable
+  include Interface::Muteable
+
+  INDICATOR = "\x2A\x53" # *S
+  HASH      = "################"
+  ERROR     = 70
```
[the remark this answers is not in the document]
caspiano: `ERROR` could be moved into the `MessageType` enum

drivers/sony/displays/bravia.cr
```diff
@@ -0,0 +1,240 @@
...
+      self[:audio_mute] = parsed_data.to_i == 1
+    when .pip?
+      self[:pip] = parsed_data.to_i == 1
+    when .volume?
+      self[:volume] = parsed_data.to_i
+    when .mac_address?
+      self[:mac_address] = parsed_data.split('#')[0]
+    when .input?
+      self[:input] = Input.from_param(parsed_data[7..15])
+    end
```
caspiano, line 238: If you implemented a `#to_symbol` on the enum, you could do something like this
```suggestion
    state = case cmd
            when .power?, .mute?, .pip?
              parsed_data.to_i? == 1
            when .volume?
              parsed_data.to_i?
            when .mac_address?
              parsed_data.split('#').first?
            when .input?
              Input.from_param(parsed_data[7..15])
            end
    self[cmd.to_symbol] = state unless state.nil?
```

caspiano: I think things are good where they are, this just reduces the code a tiny bit

caspiano: wouldn't even need a to `to_symbol` it would be more like `#to_state_key : String` because you can index into a `NamedTuple` with a string

caspiano (approved):

PR #133, opened 2021-04-07
chore(shard.lock): revert shard.lock to pre-v1 state

stakach:


kimburgess:
Noting 162631230c365b40c9ed40b3d46ce1b523196f33 as an associated change here too.

PR #134, opened 2021-04-08
build(docker-compose): remove the container_name key for redis

caspiano:
Use of the `container_name` key is not necessary here, and may lead to container name conflicts.


stakach (approved):
LGTM

PR #136, opened 2021-04-08
Sony projector

pkheav:
Only includes serial and fh drivers but not https://github.com/acaprojects/ruby-engine-drivers/blob/beta/modules/sony/projector/pj_talk.rb


stakach (changes requested):
needs to implement the Switchable interface
@tassja do you have time today to make these changes?

stakach (approved):
LGTM

PR #145, opened 2021-04-20
chore(shard.lock): bump calendar and office365

w-le:
fix event.sensitivity and fetch events limited to 10
https://github.com/PlaceOS/calendar/commit/ab8babafca01954545694da5cc116108a7af4eb3
https://github.com/PlaceOS/office365/pull/17

1. Will this cause incompatibility issues with existing PlaceOS instances that did not start off on crystal v1.0.0.?
2. Will I need to run this migration? https://github.com/PlaceOS/models/issues/75

### DIFF
--- a/shard.lock
+++ b/shard.lock
@@ -111,7 +111,7 @@ shards:
 
   office365:
     git: https://github.com/PlaceOS/office365.git
-    version: 1.12.1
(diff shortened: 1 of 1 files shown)


stakach (approved):
LGTM

PR #146, opened 2021-04-22
[WIP] signal router

### LINKED ISSUE #95: Driver Migration: signal router
**Driver to be Migrated**
Abstract signal router (https://github.com/acaprojects/ruby-engine-drivers/pull/35)

**Link to Existing Driver**
https://github.com/acaprojects/ruby-engine-drivers/blob/beta/modules/aca/router.rb

Re-implementation of concepts from graph-based signal router. The previous approach should continue to work well, however design should consider a simpler method for combining systems (and associated signal graphs) without requiring duplicate definitions.

Should be possible to read connection maps into redis keys and store the graph structure as a set of adjacency lists against these. This can then be read / updated by all instances.

kimburgess:
Provides tooling for handling signal distribution within arbitrary system configurations.

**Note: current status is work-in-progress, do not merge. PR opened for feedback and discussion.**

Closes #95, provides framework for placeos/placeos#47

### DIFF
--- a/drivers/place/rooms/meet.cr
+++ b/drivers/place/rooms/meet.cr
@@ -1,5 +1,5 @@
 require "placeos-driver"
-require "../router"
+require "placeos-driver/interface/powerable"
 
 class Place::Rooms::Meet < PlaceOS::Driver
   generic_name :System
@@ -7,19 +7,108 @@ class Place::Rooms::Meet < PlaceOS::Driver
   description <<-DESC
     Room level state and behaviours.
 
-    This driver provides a high-level API for interaction devices, systems and
-    integrations found within common workplace collaboration spaces. It's
+    This driver provides a high-level API for interaction with devices, systems
+    and integrations found within common workplace collaboration spaces. It's
     behavior will adapt to match the capabilities and configuration of other
     drivers present in the same system.
     DESC
 
   def on_load
+    load_mock_state
     on_update
   end
 
-
-  def on_update
+  enum InputType
+    Laptop
+    Wireless
+    PC
   end
 
-  include Place::Router::Core
+  enum OutputType
+    LCD
+  end
+
+  abstract class Node(Type)
+    include JSON::Serializable
+    property name : String
+    property type : Type
+    property mute : Bool? = nil
+    property n_id : UInt64 = rand(UInt64)
+  end
+
+  class Input < Node(InputType)
+    property outputs = [] of String
+    def initialize(@name, @type, @outputs); end
+  end
+
+  class Output < Node(OutputType)
+    property mod : String
+    property source : String? = nil
+    property inputs = [] of String
+    def initialize(@name, @type, @mod, @inputs); end
+  end
+
+  @inputs = {
+    "table" => Input.new("Table", InputType::Laptop, ["lcd"]),
+    "wireless" => Input.new("Vivi", InputType::Wireless, ["lcd"]),
+    "opc" => Input.new("PC", InputType::PC, ["lcd"])
+  }
+
+  @outputs = {
+    "lcd" => Output.new("LCD", OutputType::LCD, "Display_1", ["table", "wireless", "opc"])
+  }
+
+  private def update_o(name : String)
+    yield @outputs[name]
+    self["outputs/#{name}"] = @outputs[name]
+  end
+
+  private def load_mock_state
+    self[:inputs] = @inputs.map { |key, meta| self["input/#{key}"] = meta; key }
+    self[:outputs] = @outputs.map { |key, meta| self["output/#{key}"] = meta; key }
+  end
+
+  def on_update
+    self[:name] = system.name
+  end
+
+  def powerup
+    logger.debug { "Powering up" }
+    # Nothing to do...
+  end
+
+  def shutdown
+    logger.debug { "Shutting down" }
+    system.implementing(PlaceOS::Driver::Interface::Powerable).power false
+  end
+
+  def route(input : String, output : String)
+    logger.debug { "Routing #{input} -> #{output}" }
+    src = @inputs[input].n_id
+    dst = @outputs[output].n_id
+    lcd = system["Display_1"]
+    lcd.power true
+    case input
+    when "table"
+      lcd.switch_to "Hdmi"
+    when "wireless"
+      lcd.switch_to "Hdmi2"
+    when "opc"
+      lcd.switch_to "Option"
+    end
+    update_o(output) { |o| o.source = input }
+  end
+
+  def mute(input_or_output : String, state : Bool = true)
+    logger.debug { "#{state ? "Muting" : "Unmuting"} #{input_or_output}" }
+    case input_or_output
+    when "lcd"
+      system["Display_1"].mute_audio state
+      update_o("lcd") { |o| o.mute = state }
+    end
+  end
+
+  def unmute(input_or_output : String)
+    mute input_or_output, false
+  end
 end
--- a/drivers/place/rooms/meet.cr
+++ b/drivers/place/rooms/meet.cr
@@ -1,6 +1,12 @@
 require "placeos-driver"
 require "placeos-driver/interface/powerable"
 
+struct Enum
+  def to_json(json : JSON::Builder)
+    json.string(to_s.underscore)
+  end
+end
+
 class Place::Rooms::Meet < PlaceOS::Driver
   generic_name :System
   descriptive_name "Meeting room logic"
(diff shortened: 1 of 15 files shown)


kimburgess:
@stakach and @caspiano this is turning into a wall of a PR. To minimise future review effort I'd would love some early feedback.

Ignore everything in `rooms/*` at the moment, ditto with the current state of the `router` driver itself. Very keen on your thoughts of the content of [`router/*`](https://github.com/PlaceOS/drivers/tree/room-logic/drivers/place/router) before that becomes too calcified.

drivers/place/rooms/present.cr
```diff
@@ -0,0 +1,157 @@
...
+    "opc3" => Input.new("PC", InputType::PC, ["lcd3"])
+  }
+
+  @outputs = {
+    "proj" => Output.new("Projector", OutputType::LCD, "Projector_1", ["vga", "hdmi", "doccam"]),
+    "prev" => Output.new("LCD", OutputType::LCD, "Display_1", ["vga", "hdmi", "doccam"]),
+    "lcd1" => Output.new("LCD", OutputType::LCD, "Display_2", ["opc1"]),
+    "lcd2" => Output.new("LCD", OutputType::LCD, "Display_3", ["opc2"]),
+    "lcd3" => Output.new("LCD", OutputType::LCD, "Display_4", ["opc3"])
+  }
```
caspiano, line 84: i don't see this changing, any reason a NamedTuple can't be used here and above?

drivers/place/rooms/present.cr
```diff
@@ -0,0 +1,157 @@
...
+  accessor lcd2 : Display_3, implementing: InputSelection
+  accessor lcd3 : Display_4, implementing: InputSelection
+  accessor switcher : Switcher_1, implementing: Switchable
+
+  def on_load
+    load_mock_state
+    on_update
+  end
+
+  enum InputType
```
caspiano, line 35: could the logic below be batched into some sort of `Routeable` interface? or is that the basis of the `Router` class

drivers/place/router/signal_graph.cr
```diff
@@ -0,0 +1,145 @@
...
+  # not possible.
+  #
+  # Provides an `Iterator` that provides labels across each node, the edge, and
+  # subsequent node.
+  def route(source : Node::Ref, destination : Node::Ref)
+    path = g.path destination.id, source.id, invert: true
+
+    return nil unless path
+
+    path.each_cons(2, true).map do |(succ, pred)|
```
caspiano, line 131: `each_cons_pair` can be used here to save a heap allocation as it just yields the 2 elements

drivers/place/rooms/meet.cr
```diff
@@ -0,0 +1,134 @@
+require "placeos-driver"
+require "placeos-driver/interface/powerable"
+
+struct Enum
+  def to_json(json : JSON::Builder)
+    json.string(to_s.underscore)
+  end
+end
```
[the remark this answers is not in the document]
kimburgess: Ignore everything in this and `present.cr`. These are _very_ temporary to satisfy immediate needs.

drivers/place/rooms/meet.cr
```diff
@@ -0,0 +1,134 @@
...
+    logger.debug { "Shutting down" }
+    system.implementing(PlaceOS::Driver::Interface::Powerable).power false
+    self[:active] = false
+  end
+
+  def route(input : String, output : String)
+    logger.debug { "Routing #{input} -> #{output}" }
+    src = @inputs[input].n_id
+    dst = @outputs[output].n_id
+    lcd = system["Display_1"]
```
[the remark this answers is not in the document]
kimburgess: As above.

drivers/place/router/signal_graph.cr
kimburgess: `each_cons` is intentional as it provides an `Iterator` rather than forcing eval for the full path. The array it yield is re-used though (second param) so only hits the one allocation there.

drivers/place/router/signal_graph.cr
```diff
@@ -0,0 +1,150 @@
...
+      siggraph.insert node
+
+      # Track device IO in use for building active edges
+      inputs, outputs = mod_io[node.mod]
+      case node
+      when Input
+        inputs << node
+      when Output
+        outputs << node
+      end
```
caspiano, line 99: almost a silly suggestion on my behalf
```suggestion
    when Input  then inputs
    when Output then outputs
    end.try &.push(node)
```

drivers/place/router/signal_graph.cr
```diff
@@ -0,0 +1,150 @@
...
+    path = g.path destination.id, source.id, invert: true
+
+    return nil unless path
+
+    path.each_cons(2, true).map do |(succ, pred)|
+      {
+        g[succ],       # source
+        g[pred, succ], # edge
+        g[pred],       # next node
+      }
```
caspiano, line 136: I think a more descriptive type than a tuple would aid in use here as the positional data is pretty meaningless.
A struct or namedtuple perhaps?

drivers/place/router/digraph.cr
```diff
@@ -0,0 +1,196 @@
+# Labelled digraph. Holds node labels of type *N* and edge labels of type *E*.
```
caspiano, line 1: This is neat. Are you planning on extracting this to a shard after a bit of battle testing?

drivers/place/router/digraph.cr
```diff
@@ -0,0 +1,196 @@
...
+  def [](pred_id, succ_id)
+    fetch(pred_id, succ_id) do
+      raise Error.new "Edge #{pred_id} -> #{succ_id} does not exist"
+    end
+  end
+
+  # :ditto:
+  def fetch(pred_id, succ_id, &) : E
+    succ_id = check_node_exists succ_id
+    node(pred_id).succ.fetch(succ_id) { yield pred_id, succ_id }
```
caspiano, line 77: Double-check that this is yielding a tuple, there was an issue where this syntax would end up yielding twice. Thought your tests probably cover this.

drivers/place/router.cr
```diff
@@ -0,0 +1,87 @@
...
+  description <<-DESC
+    A universal matrix switcher.
+    DESC
+
+  def on_load
+    on_update
+  end
+
+  def on_update
+    # TODO: load connections config
```
caspiano, line 20: I assume this is going to be utilising `from_json` and adding edges to the `@siggraph` in `Router::Core`?

drivers/place/router/digraph.cr
```diff
@@ -0,0 +1,196 @@
...
+
+  # Perform a breadth first search across the graph, starting at *from*.
+  #
+  # Each node id is yielded as it's traversed. The search will terminate when
+  # this block returns true. If `nil` is returned the node is skipped, but the
+  # traversal continues.
+  #
+  # Results are provides as a Hash that includes all reaches nodes as the keys,
+  # and their predecessor as the associated value.
+  def bfs(from, & : UInt64 -> Bool?)
```
caspiano, line 106: I think a full name is better here. `breadth_first_search`

caspiano (commented):
Looking excellent.

drivers/place/router/digraph.cr
kimburgess: Sort of - this is an interface test. Medium term intent is to swap this out for https://git.sr.ht/~kb/sparse

drivers/place/router.cr
kimburgess: Yep correct.

drivers/place/router/digraph.cr
kimburgess: Do you have an example of that oddity in action? Will triple check things here - but keen to grok why that could happen for general language understanding.

kimburgess: Bump @caspiano

drivers/place/router/signal_graph.cr
kimburgess: Reason for this is to enable:
https://github.com/PlaceOS/drivers/blob/7ff6e942e79ac0853abb67fc53d7b0a35a24685b/drivers/place/router/signal_graph_spec.cr#L118
So that you can keep usage of this very clean.

kimburgess:
Thank you for the early discussion on internals. Closing this without merge as I've just done some history shuffling to remove the room logic sketches.

Clean branch with only router components available at https://github.com/PlaceOS/drivers/tree/signal-router. Will pop open a fresh PR when this is ready to go.

kimburgess:
Looks like this is another one that's been reopened by external tooling @jeremyw24. Active is in #209.

PR #148, opened 2021-04-26
feat: add generic MQTT driver

stakach:
allows for a single connection to be used for multiple drivers
once configured, various drivers can subscribe to interested state via typical PlaceOS subscription model


drivers/place/mqtt_spec.cr
```diff
@@ -0,0 +1,70 @@
+require "mqtt"
+
+DriverSpecs.mock_driver "Place::Pinger" do
```
caspiano, line 3:
```suggestion
 DriverSpecs.mock_driver "Place::MQTT" do
 ```

PR #149, opened 2021-04-28
style: define base ameba config and fix lints across drivers

### LINKED ISSUE #117: Define lints
To ensure drivers uniformly meet style guidelines, we use [ameba](https://github.com/crystal-ameba/ameba).
Some issues are unavoidable, or at least difficult to mitigate, like cyclomatic complexity. However other issues like unused variables and shadowing can indicate bugs or even dead code.

I suggest that a config is generated with `ameba --gen-config` and refined via a PR to a set of standards agreed on by the PlaceOS dev team.
After agreement on this config, drivers that do not pass this lint suite should not be merged.

caspiano:
The team's opinion on the contents of `.ameba.yml` would be highly appreciated.

This configuration will evolve with time, with the config and the GitHub action, the intent is to reduce review effort spent on code style.

Closes #117


stakach (dismissed):
Sweet, looks good to me

PR #153, opened 2021-04-30
ci: move to github 

caspiano:
- Remove travis environment
- Add github action based CI
- Mount `.git` to fix failing server specs
- Fix up `Dockerfile` to improve layer cacheability


kimburgess (dismissed):
Sorry this took so long to look at!

There's a lot that we can do to improve some of this setup, but may be simpler once the new build service is in place. Current iteration is a move in the right direction though - LGTM.

caspiano:
@kimburgess in lieu of separating out the test runner from this repository, would you recommend adding a cache layer to the action (I'd like to compile the binaries in release mode)? 
I would also like to run the test runner specs in parallel with the driver report. 
Any pointers are appreciated

kimburgess:
@caspiano yep, using [actions/cache](https://github.com/actions/cache) is definitely worthwhile. For it to play nicely, you'll need to use buildkit. You can do this via the [docker/build-push-action](https://github.com/docker/build-push-action), or if you want to keep things in compose, by using compose under it's new home in the `docker` cli (rather than `docker-compose`) which will let you use the `DOCKER_BUILDKIT` env var (https://docs.docker.com/compose/reference/build/#native-build-using-the-docker-cli).

Example with the action can be seen in https://github.com/PlaceOS/PlaceOS/blob/967f20c161b3179dc5489be099a49cacad17e9d1/.github/workflows/build.yml#L98-L99

PR #156, opened 2021-05-10
Gallagher rest api

stakach:
port of the ruby version


stakach:
complete in #156

w-le (approved):
Looks like it should work and would be ready for initial integration testing

PR #161, opened 2021-05-17
feat(smtp): add setting to ignore cert validation

pkheav:
Ideally it would have been better to add a setting called `setting?(OpenSSL::SSL::VerifyMode, :ssl_verify_mode)` so we could specify other enums/values of `OpenSSL::SSL::VerifyMode` instead of `setting?(Bool, :ssl_verify_ignore)` but that doesn't work because of this issue https://github.com/crystal-lang/crystal/pull/7382


caspiano (approved):

pkheav:
Thanks for approving @caspiano :)

PR #170, opened 2021-05-24
fix: clear schedules on connect

stakach:
to test if disconnect is not being called


kimburgess (dismissed):
LGTM. Schedules are only used as poll task in drivers included. Safe to merge with no functional impact.

viv-4:
Added Biamp Nexia

viv-4 (commented):
May also need to biamp nexia

viv-4 (approved):

PR #171, opened 2021-05-24
lint(mulesoft): fix ameba lints

caspiano:
- Tiny fixes for lints.
- Adds a ignore rule for `repositories`, which are external to PlaceOS/drivers

I have not addressed the comments raised in https://github.com/PlaceOS/drivers/commit/d97bb43f96ce9fc7a5cff2f174f597024e026d07 as I believe they should be checked by the implementer first.

### DIFF
--- a/.ameba.yml
+++ b/.ameba.yml
@@ -1,3 +1,6 @@
+Excluded:
+  - repositories/**/*.cr
+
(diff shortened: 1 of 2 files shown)


w-le (approved):
oh damn I should have merged this in before creating #173 , but all good I'll merge this now and then resolve any conflicts that arise in #173

PR #173, opened 2021-05-24
fix(mulesoft/exporter): status vars, perf, style, format

w-le:
Thanks for these [recommendations](https://github.com/PlaceOS/drivers/commit/d97bb43f96ce9fc7a5cff2f174f597024e026d07#comments)! The changes here should address them. Do let me know if anything else.

Apologies for committing directly to main branch instead of a PR from feat>main. I was in the middle of a client deployment change window and realised that the UAT instance was running off my fork of drivers while the PROD instance was on placeos/drivers. Will make sure to have new drivers reviewed in a PR PRIOR to deployments in the future.


drivers/mulesoft/calendar_exporter.cr
```diff
@@ -72,9 +74,8 @@ class MuleSoft::CalendarExporter < PlaceOS::Driver
   end
 
   protected def export_booking(booking : Hash(String, Int64 | String | Nil))
-    # Mulesoft booking titles are often nil. Use the body instead in this case
-    booking["title"] = booking["body"] if booking["title"].nil?
-    booking["title"] = "#{booking["recurring_master_id"]} #{booking["title"]}"
+    # Add the course code to the front of the booking title/body
+    booking["title"] = "#{booking["recurring_master_id"]} #{booking["title"] || booking["body"]}"
```
caspiano, line 78: Is there a chance that this method could be called twice for the same booking?

caspiano (dismissed):

drivers/mulesoft/calendar_exporter.cr
w-le: I believe export_booking will run once for each item in @bookings, and only AFTER @bookings is successfully updated with the (real) booking data from the Mulesoft module.

So our local booking["title"] gets transformed but then replaced with Mulesoft (no transformed) data before export_booking is run again.

If @bookings fails to get replaced with the Mulesoft bookings then I think there would be an error before hitting `@bookings.each { |b| export_booking(b) }` on line 43

drivers/mulesoft/calendar_exporter.cr
```diff
@@ -34,12 +34,13 @@ class MuleSoft::CalendarExporter < PlaceOS::Driver
 
     @time_zone_string = setting?(String, :calendar_time_zone).presence
     @time_zone = Time::Location.load(@time_zone_string.not_nil!) if @time_zone_string
-    self[:timezone] = Time.local.to_s
+    self[:timezone] = @time_zone.to_s
 
     subscription = system.subscribe(:Bookings_1, :bookings) do |_subscription, mulesoft_bookings|
-      logger.debug { "DETECTED changed in Mulesoft Bookings.." }
+      logger.debug { "DETECTED changed in Mulesoft Bookings..." }
```
caspiano, line 40:
```suggestion
      logger.debug { "DETECTED changes in Mulesoft Bookings..." }
```

drivers/mulesoft/calendar_exporter.cr
caspiano: My worry is that this method call won't be idempotent (different if run again with the same inputs), but following the code it seems okay to me.

w-le: yeah makes sense, is the best thing to do usually to clone the object before transforming, like this?

w-le: i've made a change where a booking clone is transformed instead of the booking in the instance variable. have requested new review

drivers/mulesoft/calendar_exporter.cr
```diff
@@ -34,12 +34,13 @@ class MuleSoft::CalendarExporter < PlaceOS::Driver
 
     @time_zone_string = setting?(String, :calendar_time_zone).presence
     @time_zone = Time::Location.load(@time_zone_string.not_nil!) if @time_zone_string
-    self[:timezone] = Time.local.to_s
+    self[:timezone] = @time_zone.to_s
 
     subscription = system.subscribe(:Bookings_1, :bookings) do |_subscription, mulesoft_bookings|
-      logger.debug { "DETECTED changed in Mulesoft Bookings.." }
+      logger.debug { "DETECTED change in Mulesoft Bookings..." }
       @bookings = Array(Hash(String, Int64 | String | Nil)).from_json(mulesoft_bookings)
```
caspiano, line 41: Doing something along these lines instead of the `Hash` would be much easier to maintain, faster, and ensures correctness

```crystal
record MulesoftBooking,
  getter title : String?,
  getter recurring_master_id : Int64,
  getter event_start : Int64,
  getter event_end : Int64,
  getter body : String do

  include JSON::Serializable

  def to_place_calendar(
    user_id : String,
    attendees : Array(String),
    timezone : String,
    system_email : String,
    system_name : String,
  )
    {
      title:       "#{recurring_master_id} #{title}",
      event_start: event_start,
      event_end:   event_end,
      timezone:    timezone,
      description: body,
      user_id:     system_email,
      location:    system_name,
      attendees:   [{email: system_email, name: system_name}],
    }
  end
end
```

drivers/mulesoft/calendar_exporter.cr
```diff
@@ -88,28 +91,25 @@ class MuleSoft::CalendarExporter < PlaceOS::Driver
         attendees:   [@just_this_system],
         location:    system.name.not_nil!,
       }
-      logger.debug { ">>> EXPORTING booking #{new_event}" }
+      logger.debug { ">>> EXPORTING booking as: #{new_event}" }
```
caspiano, line 94:
```crystal
protected def export_booking(booking : MulesoftBooking)
  unless event_already_exists?(booking, @existing_events)
    new_event = booking.to_place_calendar(@time_zone_string, system.email, system.name)
    logger.debug { ">>> EXPORTING booking as: #{new_event}" }
    calendar.create(**new_event)
  end
end
```

drivers/mulesoft/calendar_exporter.cr
```diff
@@ -88,28 +91,25 @@ class MuleSoft::CalendarExporter < PlaceOS::Driver
...
-    false
+  protected def event_already_exists?(booking : Hash(String, Int64 | String | Nil), existing_events : Array(JSON::Any))
+    existing_events.any? { |existing_event| booking_matches_event?(booking, existing_event.as_h) }
   end
 
-  protected def events_match?(event_a : Hash(String, Int64 | String | Nil), event_b : Hash(String, JSON::Any))
-    event_a.select("event_start", "event_end", "title") == event_b.select("event_start", "event_end", "title")
+  protected def booking_matches_event?(booking : Hash(String, Int64 | String | Nil), event : Hash(String, JSON::Any))
+    booking.select("event_start", "event_end", "title") == event.select("event_start", "event_end", "title")
   end
```
caspiano, line 105:
```crystal
protected def event_already_exists?(booking : MulesoftBooking, existing_events : Array(JSON::Any))
    logger.debug { "Checking for existing events that match: #{booking.resolved_title}" }
    existing_events.any? { |event| booking.matches?(event) }
end
```

drivers/mulesoft/calendar_exporter.cr
w-le: Thanks so much, I'll grok this after some other tasks, then see if anything else should change and then test.
Yeah I was originally intending the booking hash and calendar event to be the same object but couldn't get that to work just due to my crystal inexperience. Ran out of time due to a delivery deadline so just implemented it this way.
Now that I have time it'll be great (and very educational for me) to implement it properly.

caspiano:
@w-le I can whack in the suggestions I made if that's good for you?

PR #175, opened 2021-05-28
[WIP] adding update metadata method to StaffAPI

cacique-coder:
Hi Guys,

This is a method I am needing in Staff API.

However there are a few problems with it:

Data type are the same as the method APIWrapper::Metadata, however is not compiling using the Spec Runner.

### DIFF
--- a/drivers/place/staff_api.cr
+++ b/drivers/place/staff_api.cr
@@ -153,6 +153,15 @@ class Place::StaffAPI < PlaceOS::Driver
     placeos_client.metadata.children(id, key)
   end
 
+  def update_metadata(
+    id : String,
+    name : String,
+    details : JSON::Any | Hash | NamedTuple | Array,
+    description : String? = nil
+  )
+    placeos_client.metadata.update(id, name, details, description)
+  end
+
   # ===================================
   # ZONE INFORMATION
   # ===================================


stakach:
The issue is that `details : JSON::Any | Hash | NamedTuple | Array` is too generic for JSON serialisation.
Inside crystal you can limit something to be an Array or NamedTuple and then the compiler will do the rest of the work to check any Array(Types) you send to function are valid at compile time.

However given this function will be called from another driver the input is always JSON so the following change should resolve the issue:

```crystal
  def update_metadata(
        id : String,
        name : String,
        details : JSON::Any,
        description : String? = nil
      )
        placeos_client.metadata.update(id, name, details, description)
end
```

cacique-coder:
Hi @stakach Something interesting about this is in the repository [crystal-client](https://github.com/PlaceOS/crystal-client/blob/master/src/placeos/api_wrapper/metadata.cr#L19) it works with the same definition, I was thinking to keep the same interface between the driver and the client.

however, our first implementation in our side we were using JSON::Any and it was working correctly until a few days ago. However now we are receiving a new error during compilation.

```
Error: no overload matches 'Hash(String, String)#merge' with type NamedTuple(title: String)
 
Overloads are:
 - Hash(K, V)#merge(other : Hash(L, W))
 - Hash(K, V)#merge(other : Hash(L, W), &block)
```
image

If we replace it from `JSON::Any` to `String` or `Array(String)` <last one it was just an experiment>  and internally using JSON.parse(details) it compiles perfectly.

stakach:
fix is in [here](https://github.com/PlaceOS/driver/pull/86), once merged will be resolved with a `shards update`

stakach:
the issue is actually related to driver metadata that is being generated, JSON schema is relatively new and is generated for inter-driver communication allowing for deep introspection.

Although for the most part you'll see it when using backoffice, it'll allow for validatation of params entered into execute

stakach:
OK that is merged and after a `shards update --ignore-crystal-version` I can compile the driver
thanks for pointing this out!

PR #198, opened 2021-07-05
fix: change conflicting method name

### LINKED ISSUE #191: Lenel OpenAcess driver fails to compile
[See here](https://github.com/PlaceOS/drivers/runs/2932347703?check_suite_focus=true#step:3:3059) for compilation output. 

Looks like a slight signature mismatch.

tassja:
`before_request` was referencing the definition introduced in this [commit](https://github.com/PlaceOS/driver/commit/d416e53cb9de49d4e5916bfc889567ae5d2c7c72#diff-933bebc00e20ad0f8e3b3aad8fd2b5ce20fccd576e8ab27ba823f8109961c6f9)

Closes #191

### DIFF
--- a/drivers/lenel/open_access/client.cr
+++ b/drivers/lenel/open_access/client.cr
@@ -20,7 +20,7 @@ class Lenel::OpenAccess::Client
   property token : String?
 
   def initialize(@transport, @app_id)
-    transport.before_request do |req|
+    transport.before_lenel_request do |req|
       req.headers["Application-Id"] = app_id
       req.headers["Content-Type"] = "application/json"
       req.headers["Session-Token"] = token.not_nil! unless token.nil?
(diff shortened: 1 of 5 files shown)


stakach (changes requested):
there is a new callback called `before_request` built into the drivers now
this is implemented directly on the driver, much like the `connected` and `on_load` callbacks

So we no longer need to re-open the transport base class to do this
create a function called:

```crystal
def before_request(request : HTTP::Request)
   # ...
end
```

drivers/lenel/open_access/client.cr
```diff
@@ -20,7 +20,7 @@ class Lenel::OpenAccess::Client
   property token : String?
 
   def initialize(@transport, @app_id)
-    transport.before_request do |req|
+    transport.before_lenel_request do |req|
```
stakach, line 23: this is where the error would be as `transport` no longer has this method

stakach (changes requested):

stakach:
So I've made a change to something that will work in production but is not working with the spec.
I think I have to make an additional change to `placeos-driver` to have it function correctly in specs

stakach:
OK I think this is good to merge now @tassja 
was a little more complex than it should have been

stakach (approved):

PR #203, opened 2021-07-12
fix(officernd): refactor request functions

### LINKED ISSUE #190: OfficeRnD API driver does not compile
[See here](https://github.com/PlaceOS/drivers/runs/2932347703?check_suite_focus=true#step:3:616) for the compilation output.

Should be a simple fix. It looks like there's some request logic that could be refactored into using `responsible`

tassja:
Closes #190


stakach (approved):
LGTM

PR #209, opened 2021-07-28
Signal router

kimburgess:
Provides a virtual matrix switcher for arbitrary system designs. Functions as a standalone module for signal distribution only, or can be included in drivers which extend this for full room interaction.

Settings define physical connectivity between devices, with all other operation inferred from interfaces.

```yaml
Switcher_1:
  - *Foo
  - *Bar
Display_1:
  hdmi: Switcher_1.1
  hdmi2: *Baz
```

This is parsed into a graph representing all of the possible system connectivity with device interactions required to realise this stored on it's edges. Background info and design of this: https://docs.google.com/document/d/1DG2s9jjMVhiW65YGPDkUnOYYpFDeW42SkGvHyp1BEjQ/edit

Signal distribution is then performed with a single 'route' request.

PR #210, opened 2021-07-28
Meeting room logic

kimburgess:
Base implementation of `meet` style room.

Extends from the core signal router in #209.


caspiano (approved):
Looks perfect to me.

PR #212, opened 2021-07-30
refactor: remove spec sleeps

caspiano:
There were a bunch of `sleep`s littered through specs.
I think we should try to avoid these entirely.
if they are necessary,  we should implement some kind of asynchronous abstraction or make requests synchronous.


kimburgess:
Responses from `exec`'s can be awaiting by calling `#get` on their result, which blocks until the Future resolves. One that I've defined run into (and had to insert a sleep to accomodate) is the initial init / settings load.

This change is definitely the correct thing to do, but will need to be synchronised with improvements to spec-runner.

PR #228, opened 2021-08-10
fix(area management): area summary values should be rounded

stakach:


w-le (approved):
yep that should work. global rounding is fine for now.
a future project will probably require the global rounding to be optionally overridden only for specific values. We can deal with that when it's required.

PR #233, opened 2021-08-12
Router / room logic

kimburgess:
Added:
- **(breaking)** `PlaceOS::Meet` now implements the [`Powerable` interface](https://github.com/PlaceOS/driver/blob/master/src/placeos-driver/interface/powerable.cr) for system power.
- **(breaking)** `PlaceOS::Meet` now implements the [`Muteable` interface](https://github.com/PlaceOS/driver/blob/master/src/placeos-driver/interface/muteable.cr) for audio and video muting.
- Nicer error messages. When specifying an unknown input/output known refs are searched for a close match and suggested.

Changed:
- **(breaking)** Parameter order of `PlaceOS::Meet#volume` flipped to align with `mute` interfaces.
- Muting moved out of `PlaceOS::Router::Core` and into room logic.

Deprecated:
- `PlaceOS::Meet#powerup` - use `power(true)`
- `PlaceOS::Meet#shutdown` - use `power(false)`

Fixed:
- Incorrect event ordering that would cause outputs to attempt to switch input before powering on.

### DIFF
--- a/drivers/place/meet.cr
+++ b/drivers/place/meet.cr
@@ -47,7 +47,7 @@ class Place::Meet < PlaceOS::Driver
   end
 
   # Set the volume of a signal node within the system.
-  def volume(input_or_output : String, level : Int64)
+  def volume(level : Int32, input_or_output : String)
     logger.info { "setting volume on #{input_or_output} to #{level}" }
     node = signal_node input_or_output
     node.proxy.volume level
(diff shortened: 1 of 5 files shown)


caspiano (approved):
Great. The changelog was helpful as too.
The levenshtein addition to router is awesome.

PR #234, opened 2021-08-15
Lenel Openaccess: Support querying Logged Events

w-le:
- Support querying logged events that match a given filter
- Support querying logged events within a given Time range, that match a given filter
- tidy up timezones implemention

### DIFF
--- a/drivers/lenel/open_access.cr
+++ b/drivers/lenel/open_access.cr
@@ -170,13 +170,10 @@ class Lenel::OpenAccess < PlaceOS::Driver
     personid : Int32,
     activate_epoch : Int32,
     deactivate_epoch : Int32,
-    uselimit : Int32? = nil,
-    timezone : String? = nil
+    uselimit : Int32? = nil
   )
-    time_zone = timezone || default_timezone
-    activate = Time.unix(activate_epoch).in Time::Location.load(time_zone)
-    deactivate = Time.unix(deactivate_epoch).in Time::Location.load(time_zone)
-    logger.debug { "Creating badge for cardholder #{personid}, valid from: #{activate} til #{deactivate}" }
+    activate = Time.unix(activate_epoch).in Time::Location.load(default_timezone)
+    deactivate = Time.unix(deactivate_epoch).in Time::Location.load(default_timezone)
 
     create_badge(
       type: type,
@@ -206,13 +203,10 @@ class Lenel::OpenAccess < PlaceOS::Driver
     activate_epoch : Int32,
     deactivate_epoch : Int32,
     id : Int64? = nil,
-    uselimit : Int32? = nil,
-    timezone : String? = nil
+    uselimit : Int32? = nil
   )
-    time_zone = timezone || default_timezone
-    activate = Time.unix(activate_epoch).in Time::Location.load(time_zone)
-    deactivate = Time.unix(deactivate_epoch).in Time::Location.load(time_zone)
-    logger.debug { "Updating badge #{badgekey} with id #{id} valid from: #{activate} til #{deactivate}" }
+    activate = Time.unix(activate_epoch).in Time::Location.load(default_timezone)
+    deactivate = Time.unix(deactivate_epoch).in Time::Location.load(default_timezone)
 
     update_badge(
       badgekey: badgekey,
--- a/drivers/lenel/open_access.cr
+++ b/drivers/lenel/open_access.cr
@@ -33,8 +33,9 @@ class Lenel::OpenAccess < PlaceOS::Driver
     wrapper
   end
 
-  private getter default_timezone : String do
-    setting?(String, :timezone) || "UTC"
+  private getter default_timezone : Time::Location do
+    tz = setting?(String, :timezone) || "UTC"
+    Time::Location.load tz
   end
 
   def on_load
@@ -172,8 +173,8 @@ class Lenel::OpenAccess < PlaceOS::Driver
     deactivate_epoch : Int32,
     uselimit : Int32? = nil
   )
-    activate = Time.unix(activate_epoch).in Time::Location.load(default_timezone)
-    deactivate = Time.unix(deactivate_epoch).in Time::Location.load(default_timezone)
+    activate = Time.unix(activate_epoch).in default_timezone
+    deactivate = Time.unix(deactivate_epoch).in default_timezone
 
     create_badge(
       type: type,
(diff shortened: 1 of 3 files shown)


drivers/lenel/open_access.cr
```diff
@@ -34,7 +34,7 @@ class Lenel::OpenAccess < PlaceOS::Driver
   end
 
   private getter default_timezone : String do
-    setting?(String, :timezone) || "Asia/Dubai"
+    setting?(String, :timezone) || "UTC"
   end
```
kimburgess, line 38:
```suggestion
  private getter default_timezone : Time::Location do
    tz = setting?(String, :timezone) || "UTC"
    Time::Location.load tz
  end
```

kimburgess (changes requested):
Mixing in TZ handling here feels fragile. Any parameters that accept `Time` types or models that include Time attributes will currently map from ISO 8601, which includes an offset.

Adding in the epoch based method looks good, but can likely avoid this intermediate mapping.

drivers/lenel/open_access.cr
w-le: thanks yeah that totally makes sense, i'll add it

kimburgess:
@w-le initial comment still holds. Why is the Timezone needed at all?

w-le:
yeah true it's only for those `..._epoch` versions of the create methods (intended for use with backoffice where users can only enter string/Int/bool types).
So I've made a change such that the driver will use the timezone of (in order, where present): 
`Time` param || `timezone` String param || local timezone

w-le:
Oh forgot to mention, the reason why I wanted to ensure that the _epoch methods allow the user to define the timezone is because I witnessed an Openaccess API bug where badges created/edited with a time parameter that is in a different timezone to that of the Openaccess server result in completely incorrect timestamp properties on the resulting Lenel object (e.g. 20 years in the future).

The local timezone of the driver may not always match the timezone of the openaccess server, so _epoch methods allow the user to optionally define which timezone will be used in the timestamps posted to openaccess.

kimburgess:
Times specified in epoch form by definition do not have a timezone associated. There should be no reason to ever pass a timezone along with an epoch reference.

The behaviour that you are working around instead appears to indicate a mismatch in serialisation behaviour between this driver and the OpenAccess API (or API bug), so let's address that. From the docs, expected time formats are:
647063C4-167F-4D16-82D8-8B445D86CF24

The driver does serialise to ISO 8601 time, but is worth noting UTC times will use the shorthand `Z` for the offset (e.g. `2021-08-17T06:03:18Z`). This is per ISO spec, but can't see usage of this specifically in the Lenel docs. Do you have access to the live system and ability to test this theory?

kimburgess:
> intended for use with backoffice where users can only enter string/Int/bool types

Also worth noting: [`Time` deserialises from a ISO 8601 string](https://play.crystal-lang.org/#/r/brrm). Any argument that takes a `Time` parameter can just be specified in it's string form.

w-le:
Yep I have this debug output which I posted an ISO timestamp in UTC. I've emailed a reseller to let Openaccess API team know about the issue

> Hi _____, Nothing urgent or major but maybe you'd like to forward this to Lenel's OpenAccess API team.
In my extensive testing I've found that when we create new Badges with activate/deactivate datetime (string) in UTC format they work fine as expected UNTIL there is some kind of normal API error response (e.g. this cardholder already has the maximum number of cards allowed by the system).

After those normal kind of errors, ALL subsequent API requests to create Badges that use an identical request Body format (activate/deactivate values in UTC time) will result in Badges that have unexpected activate/deactivate timestamps in the far distant future (example request body and resulting Badge properties in screenshot below). The disparity between the resulting dates and the provided dates increases by several hundred years each time there is an unrelated API error response.

Our current simple workaround is to convert the activate/deactivate datetime (string) to the __(a location)___ timezone first, even though this technically should not be required because the OpenAccess API docs state that those parameters are ISO 8601 standard.

So it's nothing major or blocking but did take me several hours of testing to confirm the issue (I initially thought it was a PlaceOS issue) and then confirm that the workaround reliably avoids the issue.

image

w-le:
So yeah I believe Openaccess is not compatible with the trailing `Z` for UTC timestamp.
I have not tried with `+00:00` instead, as I'm not sure how to make the driver change to post that

kimburgess:
Cool, that format can be switched. Quickest way to confirm is changing that serialisation for all `Time` fields.  This can be achieved with:

```crystal
struct Time
  def to_json(json : JSON::Builder)
    json.string(self.to_s "%FT%T%:z")
  end
end
```

A good spot for this would be [`open_access/models.cr`](https://github.com/PlaceOS/drivers/blob/master/drivers/lenel/open_access/models.cr#L2).

w-le:
Awesome! thanks I didn't even think about overriding that method. Should use that more often.

drivers/lenel/open_access/models.cr
```diff
@@ -1,5 +1,12 @@
 require "json"
 
+# Ensure that UTC time strings provide the offset as "+00:00" instead of "Z", as required by Openaccess
+struct Time
+  def to_json(json : JSON::Builder)
+    json.string(self.to_s "%FT%T%:z")
+  end
+end
```
caspiano, line 8: I'd advise against reopening time in case of unintended consequences elsewhere.
This is an alternative using [JSON converters](https://crystal-lang.org/api/latest/JSON/Serializable.html#usage)

```crystal
module Lenel::TimeConverter
  def self.to_json(value : Time, json : JSON::Builder)
    json.string(value.to_s("%FT%T:z")
  end
  # Can also include a `def self.from_json(json : JSON::PullParser) : Time`
end

...

# Then on a time attribute in a model
@[JSON::Field(converter: Lenel::TimeConverter)]
getter time : Time

...
```

kimburgess: The `Z` and `+00:00` forms are equivalent and both valid (for properly compliant parsers) so _should_ be safe in this specific case.

@w-le definitely follow this advice to reduce kludge factor before merge if this does solve that oddity. Likewise for anywhere else you run into that needs custom serialisation. Above is the idiomatic approach.

w-le: Thanks! implemented

drivers/lenel/open_access/client.cr
```diff
@@ -163,4 +163,28 @@ class Lenel::OpenAccess::Client
...
+    (~transport.get(
+      path: "/logged_events?version=1.0&#{params}",
+    ) >> NamedTuple(
+      page_number: Int32?,
+      page_size: Int32?,
+      total_pages: Int32,
+      total_items: Int32,
+      count: Int32,
+      item_list: Array(Event),
+    ))[:item_list]
```
caspiano, line 188: This potentially should be a model by itself, but I'm happy as is

caspiano (approved):

PR #235, opened 2021-08-19
feat(meraki/locations): Apply device OS filter to list of Locations too

w-le:
There is an existing, optional regex filter for device operating system `regex_filter_device_os`, that applied only to the `locate_user` method.
This change uses the same filter and applies it to all exposed Locations (often rendered in apps as "device dots" from AreaManagement bindings).

A common use case is if staff have a particular managed device type of interest (e.g. Windows 10 laptops, or iPhones), and the enterprise only wants to LOCATE and COUNT (e.g. real-time device dots and retrospective analytics) those specific devices, then the device os regex filter setting can be set, and `AreaMangement` driver's `duplication_factor` should be set to `1.0`


stakach (approved):
LGTM

PR #237, opened 2021-08-20
ci: run one test at a time

caspiano:
Until the driver test runner safely supports concurrent specs, we should run one spec at a time.


caspiano:
Superseded by https://github.com/PlaceOS/driver/pull/108

PR #239, opened 2021-08-23
chore(harness): crystal 1.1.1 instead of 1.0.0

w-le:
Update default driver test harness crystal version to 1.1.1

Tested briefly OK (ubuntu 20.04 in WSL2, Win10) with `./harness up` 
and compiled a driver that used Time `.step` (new in 1.1.0), which was not compiling in 1.0.0

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 services:
   # Driver test harness
   drivers:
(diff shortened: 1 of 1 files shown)


caspiano (approved):

caspiano:
One thing to double check is that core is on `1.1.1` as well

PR #244, opened 2021-09-08
chore(docs): fix crystal docs generation

stakach:
adds CI to ensure docs can always be generated


stakach:
@kimburgess added a spec for testing doc generation
but had to make some minor modifications to place/router.cr are you able to take a look at that one?

kimburgess:
Running `crystal docs` [pulls everything under `./src` into a single compiler context](https://github.com/crystal-lang/crystal/blob/b88b842d425de7ce2a74c1acf82b30c447e1b254/src/compiler/crystal/command/docs.cr#L127). This is very different to the way actual driver compilation runs, which appears to be the reason for most of these changes. Behaviour of this can be changed by passing a set of files to explicitly include. Unfortunately though due to the way glob expansion works there is no ability to neatly declare "everything but the specs" (e.g. `drivers/**/*!(_spec).cr` is _not_ valid).

Going down this current path may cause a bit of unnecessary pain and inconvenience as spec content needs to be explicitly set to `:nodoc`. More importantly, what's documented may be very different to what's actually compiled (e.g. a naming conflict across drivers re-opening a class in the docs build).

IMO this may be a good point to revisit this repo structure and make it a little more 'standard': source under `./src`, specs under equivalent path in `./spec` as this will help with compatibility with other tools too. There may also be some value in explicitly declaring driver entry points as keys under `targets` in `shard.yml` as this could then be used for discovery.

Definitely appreciate the above is a _significant_ change though and will likely impact existing systems. If something very immediate is needed for doc gen, what about just adding an `rm drivers/**/*_spec.cr` ahead of the docs build?

caspiano (dismissed):

PR #249, opened 2021-09-14
including fetching base64 pdf, some of them are coming as binary

cacique-coder:
When the pdf is a binary, we are receiving an error from the driver.


stakach (approved):
LGTM

PR #253, opened 2021-09-20
feat(vergesense/mqtt_export): New VS > MQTT logic driver

w-le:
A simple logic driver that subscribes to changes in a neighboring Vergesense driver and publishes OCCUPANCY numbers for each space (desk/room/open-area) to a neighboring Generic MQTT driver.

The Vergesense floors to be considered must be configured as setting.
The MQTT base topic must be configured as setting.
The format of the MQTT Topic to publish to is static.

Tested OK e2e on placeos-dev server.


stakach (approved):
LGTM

PR #261, opened 2021-10-05
feat(router): persist node status across reloads

### LINKED ISSUE #260: Signal router: presence signal state on reload
When reloading settings, status info associated with active routes is dropped.

To support setting changes on live systems, this should be preserved across setting reloads.

kimburgess:
Resolves #260.


stakach (approved):
LGTM

PR #262, opened 2021-10-06
feat(router): graph-based muting

kimburgess:
Adds support for graph-based mute activation.

Edges to a symbolic `MUTE` source are inserted for all devices without outputs (e.g. displays) and intermediate signal devices that implement the `Muteable` interface.
Muting may then be activated by routing from a `"MUTE"` input ref.
This will find the closest available blanking source, preferences an on-device mute, then falling back to an upstream blanking source such as that provided on a matrix switcher output.
An optional `max_dist` parameter may be specified to `route` execs to limit both this and other route operations. For example, to use on-device muting only: `route("MUTE", "foo", max_dist: 1)`.


kimburgess:
Oh, wasn't quite ready for merge :(

PR #264, opened 2021-10-07
fix(place area management): draw areas if no sensors or wireless users

stakach:


w-le (approved):
thanks

w-le:
confirmed working OK

PR #269, opened 2021-11-10
Added an ability to use null values for iaq in the sensors

grkek:


drivers/vergesense/location_service.cr
```diff
@@ -212,7 +212,8 @@ class Vergesense::LocationService < PlaceOS::Driver
             when .air_quality?
               id = "air"
               time = space.environment.try &.timestamp
-              space.environment.try &.iaq.value
+              iaq = space.environment.try &.iaq
+              iaq.nil? ? nil : iaq.not_nil!.value
```
caspiano, line 216:
```suggestion
              space.environment.try(&.iaq.try(&.value))
```

drivers/vergesense/room_sensor.cr
```diff
@@ -62,9 +62,10 @@ class Vergesense::RoomSensor < PlaceOS::Driver
       self[:people] = 0
     end
 
+    iaq = floor_space.environment.try &.iaq
```
caspiano, line 65:
```suggestion
```

drivers/vergesense/room_sensor.cr
```diff
@@ -62,9 +62,10 @@ class Vergesense::RoomSensor < PlaceOS::Driver
...
     self[:humidity] = floor_space.environment.try &.humidity.value
     self[:temperature] = floor_space.environment.try &.temperature.value
-    self[:air_quality] = floor_space.environment.try &.iaq.value
+    self[:air_quality] = iaq.nil? ? nil : iaq.not_nil!.value
```
caspiano, line 68:
```suggestion
    self[:air_quality] = floor_space.environment.try(&.iaq.try(&.value))
```

Not ideal, but maybe it's worth allowing the other sensors to be nil, and creating a helper that can warn when an integration is not returning values. Suggestion below might be a bit overkill

```crystal
module Vergesense
  struct Environment
    {% begin %}
    {% for reading in [iaq, temperature, humidity] %}
      getter {{ reading }} : Sensor?
      
      def read_{{ reading }}
        if reading = {{ reading }}
          reading.value
        else
          Log.warn { "missing reading for {{ reading }}" }
          nil
        end
      end
    {% end %}
    {% end %}
  end
end
```

drivers/vergesense/room_sensor.cr
```diff
@@ -136,7 +137,8 @@ class Vergesense::RoomSensor < PlaceOS::Driver
             when .air_quality?
               id = "air_quality"
               time = space.environment.try &.timestamp
-              space.environment.try &.iaq.value
+              iaq = space.environment.try &.iaq
+              iaq.nil? ? nil : iaq.not_nil!.value
```
caspiano, line 141: Ditto suggestion as previous iag access.

caspiano (changes requested):

drivers/vergesense/room_sensor.cr
stakach:
```suggestion
              space.environment.try(&.iaq.try(&.value))
```

PR #271, opened 2021-11-22
fix(vergesense/mqttt_export): remove root topic

pkheav:


w-le (approved):
According to the mqtt broker admin, this is apparently what is required

PR #274, opened 2021-11-24
Revert "fix(vergesense/mqttt_export): remove root topic (#271)"

pkheav:
This reverts commit 5d9a27031720a22e2792c3049c54773820687da8.


pkheav:
Sorry @w-le , I had a misunderstanding on what the root topic was and looks like we still need it but just that I was using the incorrect value before

w-le (approved):
yeah that makes more sense to me

PR #276, opened 2021-12-14
Vive leap

stakach:
new vive leap protocol driver

### DIFF
--- a/drivers/lutron/vive_leap.cr
+++ b/drivers/lutron/vive_leap.cr
@@ -1,0 +1,71 @@
+require "placeos-driver"
+require "placeos-driver/interface/sensor"
+
+class Lutron::ViveLeap < PlaceOS::Driver
+  include Interface::Sensor
+
+  # Discovery Information
+  descriptive_name "Lutron Vive LEAP"
+  generic_name :Lighting
+
+  # Requires TLS negotiation (max 10 connections)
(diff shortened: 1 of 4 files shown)


drivers/lutron/vive_leap.cr
```diff
@@ -0,0 +1,253 @@
...
+    logger.debug { "Lutron sent: #{data}" }
+    request = Request.from_json(data)
+
+    url = request["Url"]?
+    status = request["StatusCode"]? || "200 OK"
+    message_type = request["MessageBodyType"]?
+
+    # process the message based on its type by preference
+    case message_type
+    when "OneClientSettingDefinition"
```
tassja, line 153: Maybe make an enum for the MessageBodyTypes so you can do `case in...` to ensure exhaustive coverage?

tassja (dismissed):

drivers/lutron/vive_leap.cr
tassja: Style spec is also failing

tassja (commented):
Style spec is also failing

drivers/lutron/vive_leap.cr
stakach: I personally don't think it's required here as we only care about a small subset of the message body types

PR #278, opened 2022-01-12
feat: SecureOS driver

chillfox:
Should I parse the timestamp or leave it as a String?

Would it be better to combine subscribe_states and subscribe_events into a single subscribe method?

Should I create an unsubscribe method, or just hope that disconnecting clears the subscriptions?


stakach:
separating the subscribe methods is good :)
I think parsing the timestamps using a converter is a good idea (much easier to parse later)

```crystal
# something like this
@[JSON::Field(converter: Time::Format.new("%FT%T", Time::Location::UTC))]
property updated_at : Time
```

I think disconnecting would clear subscriptions.
In fact what might be useful is:

* Using the REST API to obtain a list of cameras and then auto-subscribing to them all after the connection is established
* periodically checking that the list hasn't changed and subscribing any new cameras
* having a status `self["camera_list"] = ` that has a list of all the cameras

as really we just want all the data available at all times.

chillfox:
The list of cameras is updated every 60 seconds.

All cameras are subscribed to the status changes: ["attached", "armed", "alarmed"] every 60 seconds.

All cameras are subscribed to all events every 60 seconds. This could potentially be a lot of events, so I am thinking it might be a good idea to make the event list a setting.

I am assuming that nothing bad comes from subscribing to the same camera/event more than once.

drivers/secure_os/ws_api.cr
```diff
@@ -0,0 +1,167 @@
...
+    client.basic_auth **basic_auth
+    response = client.get "#{@rest_api_host}/api/v1/cameras"
+    if response
+      json_response = RestResponse.from_json response.body
+      self["camera_list"] = @camera_list = json_response.data
+    else
+      raise "Faild to get camera list"
+    end
+  rescue error
+    logger.warn(exception: error) { "Faild to get camera list" }
```
caspiano, line 139:
```suggestion
    logger.warn(exception: error) { "Failed to get camera list" }
```

drivers/secure_os/ws_api.cr
```diff
@@ -0,0 +1,167 @@
...
+  def camera_list
+    host = setting?(Bool, :shared_host) ? config.uri.not_nil! : @rest_api_host
+    client = HTTP::Client.new URI.parse(host)
+    client.basic_auth **basic_auth
+    response = client.get "#{@rest_api_host}/api/v1/cameras"
+    if response
+      json_response = RestResponse.from_json response.body
+      self["camera_list"] = @camera_list = json_response.data
+    else
+      raise "Faild to get camera list"
```
caspiano, line 136: You're raising here, then capturing the exception immediately after.

```suggestion
      logger.error "Failed to get camera list"
```

drivers/secure_os/ws_api.cr
```diff
@@ -0,0 +1,167 @@
...
+    end
+
+    camera_list
+    subscribe_states
+    subscribe_events
+
+    schedule.every(30.seconds) { send({type: :get_server_time}.to_json, name: :server_time) }
+    schedule.every(60.seconds) { camera_list }
+    schedule.every(60.seconds) { subscribe_states }
+    schedule.every(60.seconds) { subscribe_events }
```
stakach, line 55: I would have a single block here:

```suggestion
    schedule.every(5.minutes, immediate: true) do
      camera_list
      subscribe_states
      subscribe_events
    end
```

the `immediate: true` means it'll run straight away

drivers/secure_os/ws_api.cr
```diff
@@ -0,0 +1,167 @@
...
+  end
+
+  def subscribe_events(events : Array(String)? = nil)
+    if cameras = @camera_list
+      cameras.each do |camera|
+        subscribe_events camera.id, camera.type, events
+      end
+    else
+      logger.warn { "No cameras to subscribe to" }
+    end
```
stakach, line 103: Instead of warning about no cameras I think it might be better to just do something like
```crystal
cameras = @camera_list
return unless cameras

cameras.each do |camera|
  subscribe_events camera.id, camera.type, events
end
```

as it would bombard with warnings and no cameras is probably unlikely / easy enough to discover

stakach (changes requested):
looking really good so far!

caspiano:
Formatter is producing changes, could you please run `crystal tool format`

drivers/secure_os/ws_api.cr
chillfox: I have changed it to default an empty array instead of nil.

```crystal
@camera_list : Array(Camera) = [] of Camera
```

```crystal
@camera_list.each do |camera|
  subscribe_events camera.id, camera.type, events
end
```

drivers/secure_os/ws_api.cr
```diff
@@ -0,0 +1,167 @@
...
+  end
+
+  def connected
+    host = setting?(Bool, :shared_host) ? config.uri.not_nil! : @rest_api_host
+    client = HTTP::Client.new URI.parse(host)
+    client.basic_auth **basic_auth
+    response = client.get "#{@rest_api_host}/api/v1/ws_auth"
+    if response.success?
+      json_body = JSON.parse response.body
+      token = json_body["data"]["token"].as_s
```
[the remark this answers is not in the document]
chillfox: ok, I have changed it to use JSON::Serializable

chillfox:
> Formatter is producing changes, could you please run `crystal tool format`

done

drivers/secure_os/ws_api.cr
```diff
@@ -0,0 +1,155 @@
...
+  def subscribe_events(events : Array(String)? = nil)
+    @camera_list.each do |camera|
+      subscribe_events camera.id, camera.type, events
+    end
+  end
+
+  def subscribe_events(
+    camera_id : String,
+    camera_type : String = "LPR_CAM",
+    events : Array(String)? = ["CAR_LP_RECOGNIZED"]
```
caspiano, line 97: An enum for `camera_type` and an enum for `event` will make interacting with the driver safer

chillfox: I haven't seen a list of possible camera types or event types.
The states however should be an exhaustive list ["attached", "armed", "alarmed"], so I will change that one.

drivers/secure_os/ws_api.cr
```diff
@@ -58,7 +58,9 @@ class SecureOS::WsApi < PlaceOS::Driver
     schedule.clear
   end
 
-  def subscribe_states(states : Array(String) = ["attached", "armed", "alarmed"])
+  def subscribe_states(
```
stakach, line 61: This method should be protected as it shouldn't be called from another module
Also worth noting that drivers don't allow overloading of public function names - as you can't differentiate them when making remote procedure calls

chillfox: with auto subscribing to everything I guess we probably don't need any of the subscribe methods exposed

stakach: works for me

chillfox: We can replace all the subscription methods with a single subscribe method that takes the states/events to subscribe to. And then use a setting to allow changing the states/events or leave the settings out in case it's not needed.

```crystal
private def subscribe_all(
    states : Array(StateType)? = [StateType::Attached, StateType::Armed, StateType::Alarmed],
    events : Array(String)? = nil
  )
    rules = [] of NamedTuple(
      type: String,
      id: String,
      states: Array(StateType)?,
      events: Array(String)?,
      action: Symbol,
      )

    @camera_list.each do |camera|
      rules << {
        type: camera.type,
        id: camera.id,
        states: states,
        events: nil,
        action: :STATE_CHANGED,
      }
      rules << {
        type: camera.type,
        id: camera.id,
        states: nil,
        events: events,
        action: :EVENT,
      }
    end

    return if rules.empty?

    send({
      type: :subscribe,
      # id: 1234, # optional id used in error responses
      data: {
        add_rules: rules,
      },
    }.to_json, wait: false)
  end
```

chillfox:
I have combined the subscribe methods into a single one that subscribes to events/states for all cameras in a single request.

I have also tried to set up the settings to make the stats/events subscribed to configurable. Not really sure if it would be better to default them to nil instead of an empty array.

chillfox:
Added methods to add/remove licence plates from/to watchlists

drivers/secure_os/ws_api.cr
```diff
@@ -0,0 +1,204 @@
...
+  default_settings({
+    shared_host:   true,
+    rest_api_host: "http://172.16.1.120:8888",
+    basic_auth:    {
+      username: "srvc_acct",
+      password: "password!",
+    },
+    camera_types:  [] of String,
+    camera_states: [StateType::Attached, StateType::Armed, StateType::Alarmed],
+    camera_events: [] of String,
```
chillfox, line 20: camera_types specifies which camera types to subscribe to, if it is empty then all cameras will be subscribed to.
Same thing for camera_states and camera_events.
We could default states and events to the few examples provided in the docs?

chillfox:
There doesn't appear to be any way of getting a list of license plates on watch lists. But it is possible to check if a license plate is on a watch list. Would it be worth implementing a method for that?

drivers/secure_os/ws_api.cr
```diff
@@ -0,0 +1,204 @@
...
+    @camera_types = setting Array(String), :camera_types
+    @camera_states = setting Array(StateType), :camera_states
+    @camera_events = setting Array(String), :camera_events
+  end
+
+  def connected
+    host = setting?(Bool, :shared_host) ? config.uri.not_nil! : @rest_api_host
+    client = HTTP::Client.new URI.parse(host)
+    client.basic_auth **basic_auth
+    response = client.get "#{@rest_api_host}/api/v1/ws_auth"
```
stakach, line 48: this code is repeated a few times, work creating a function so it's not repeated

```crystal
protected def http_client
  host = setting?(Bool, :shared_host) ? config.uri.not_nil! : @rest_api_host
  client = HTTP::Client.new URI.parse(host)
  client.basic_auth **basic_auth
  client
end
```

Then can do `response = http_client.get "#{@rest_api_host}/api/v1/ws_auth"`

stakach (changes requested):

drivers/secure_os/ws_api.cr
chillfox: Yeah, that one kinda snuck up on me with more and more get/post requests slowly getting added.

stakach (approved):
LGTM

caspiano (approved):
Great work!

PR #279, opened 2022-01-12
Fixing property type on Floorsense::UserGroup

cacique-coder:
There is a typo not he property count, it should be Int32 with Capital I


stakach (approved):
thanks!

PR #281, opened 2022-02-16
Improvements/floorsense reservation lockers

cacique-coder:
Expanding current driver with the following features

1. controller_list: can accept 2 parameters to be able to filter controllers if the have lockers, desks or booths.
2. method to get an individual reservation: Floorsense has a limitation which can't pull every reservation existing. So during sync it is possible not getting the reservation from Floorsense when query all of them, But we can know the id in specific to be requested.
3. the action res-list, also accepts cid parameter, which can help to pull reservations for a specific controller.


stakach (approved):
lgtm

PR #283, opened 2022-02-24
feat: add Webex InstantConnect driver

tassja:
Closes #283


drivers/webex/instant_connect.cr
```diff
@@ -0,0 +1,37 @@
...
+  # Discovery Information
+  generic_name :InstantConnect
+  descriptive_name "Webex InstantConnect"
+  uri_base "https://mtg-broker-a.wbx2.com"
+
+  default_settings({
+    bot_access_token: "token",
+  })
+
+  @audience_setting = ""
```
stakach, line 11: missing the extraction of the audience settings
the default setting should be: `a4d886b0-979f-4e2c-a958-3e8c14605e51` (which is CISCOs one and we won't need to change often)

stakach (changes requested):

drivers/webex/instant_connect.cr
tassja: Ah ok I wasn't sure if that was secret

PR #285, opened 2022-02-28
Add Webex bot drivers

grkek:


stakach (changes requested):
As discussed, remove the custom websocket implementation
the websocket is established during class initialisation so no need for `on_open`

```
@socket = socket = Websocket.new
socket.on_message ...
socket.on_binary ...
socket.send message.to_json
socket.run
```

stakach:
@grkek made the changes to the websocket
what else needs to happen with this one? is it ready for testing?

grkek:
> @grkek made the changes to the websocket what else needs to happen with this one? is it ready for testing?

We need to add a command with which we will be able to register events, for example:

A booking event, a guest arriving event, etc.

stakach (approved):
LGTM

PR #290, opened 2022-03-14
adding desk_scan and activate booking into floorsense driver

cacique-coder:
Guy we need to add active booking because when you are arriving early to the desk, confirm booking fails.
But using active booking endpoint allows to use the desk!


w-le:
yep Phil talked to us about this, do you have the new floorsense API endpoint and params that we should add to our driver? We will make the driver use this new endpoint when current time < time of booking.

stakach (approved):
LGTM, thanks!

PR #300, opened 2022-04-06
feat: port QSC Q-SYS Camera ruby driver

rdiazx:
**Description of the change**

Ported existing ruby driver code to crystal.

**Note - TODO**

I didn't add the spec test for this driver. It depends on a driver with generic_name:Mixer - like the [q_sys_control.cr](https://github.com/PlaceOS/drivers/blob/chore/port-ruby-driver/drivers/qsc/q_sys_control.cr) - and I'm not sure of the expected return values.


stakach (approved):
LGTM

stakach (changes requested):
Actually, I think we should implement the other common camera interfaces
https://github.com/PlaceOS/drivers/blob/master/drivers/sony/camera/cgi_protocol.cr#L8

This implements a bunch of other interfaces
https://github.com/PlaceOS/driver/blob/master/src/placeos-driver/interface/camera.cr#L6

If there is a function that you can't implement due to limitations of what this camera can do, then just have it raise an error

drivers/qsc/q_sys_camera.cr
```diff
@@ -0,0 +1,92 @@
+# encoding: ASCII-8BIT
+# frozen_string_literal: true
```
pkheav, line 2: forgot to mention this the other day @rdiazx  but I don't think these lines are needed in PlaceOS, can you confirm @stakach ?

stakach: Yeah not needed, all strings are UTF8 in crystal
That was a ruby hack to get around windows encoding leaking in if people edited files on windows

PR #302, opened 2022-04-13
Add missing fields to the user model

Hamada92:
This PR adds some missing fields to the user model in floorsense.

### DIFF
--- a/drivers/floorsense/models.cr
+++ b/drivers/floorsense/models.cr
@@ -265,6 +265,14 @@ module Floorsense
     property desc : String?
     property lastlogin : Int64?
     property expiry : Int64?
+    property reslimit : Int64?
+    property pin : Int64?
+    property ugroupid : Int64?
+    property uidtoken : String?
+    property extid : String?
+    property usertype : String?
+    property desc : String?
+    property privacy : String?
   end
 
   class LogEntry


w-le (approved):
the params aligns, apart from `privacy` which isn't in the floorsense docs (maybe it's new?)
https://apiguide.smartalock.com/#2b73222c-a86e-4cc2-8a12-d8d307adf3b1

Example Response from floorsense api
```
{
  "result": true,
  "type": "response",
  "info": {
    "email": "simon@amixcapital.com",
    "uid": "31344311",
    "usertype": "user",
    "pin": "8721",
    "name": "Simon Hope",
    "lastlogin": 1582692563,
    "expiry": 0,
    "extid": "9876543210987654321",
    "reslimit": 0,
    "uidtoken": "694433b85d6c7b2dd2c37c75a9c9a401",
    "desc": "#firstaid"
  }
}
```

Hamada92:
Thanks @w-le  , if we keep the privacy one there, would it break anything?

w-le:
yep I don't think it will break anything so I've approved the PR. We'll let @stakach review as well since he is much more familiar with this driver.

stakach (approved):
LGTM

PR #305, opened 2022-04-20
Fix: user pin string

Hamada92:
Hi guys, last week I added these fields, but there is a bug, pin is a string.


w-le:
yep thanks, looks like User.privacy is an Int too, based on floorsense api docs example response for GET user-list

w-le (approved):

PR #322, opened 2022-05-11
Limit offset user list

Hamada92:
Hi guys, this PR adds a limit/offset to the user_list method in floorsense.
Can I have a review plese?


stakach (approved):
LGTM

PR #325, opened 2022-05-12
Update protocol3000.cr

yarah123:
On line 2, I added the 'require' statement to 'include' the switchable interface on line 55.


stakach:
Thanks, we definitely need to add this interface to this driver, partly why it isn't merged yet

PR #330, opened 2022-05-30
adding signal method to staff_api

cacique-coder:
**Description of the change**

Adding signal method.

### DIFF
--- a/drivers/place/staff_api.cr
+++ b/drivers/place/staff_api.cr
@@ -532,7 +532,7 @@ class Place::StaffAPI < PlaceOS::Driver
   end
 
   @[Security(Level::Support)]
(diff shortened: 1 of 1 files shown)


drivers/place/staff_api.cr
```diff
@@ -531,6 +531,11 @@ class Place::StaffAPI < PlaceOS::Driver
     JSON.parse(response.body)
   end
 
+  @[Security(Level::Support)]
+  def signal(channel : String, payload : JSON::Any)
```
caspiano, line 535: When signalling a channel, you should be able to pass an empty payload.

```suggestion
  def signal(channel : String, payload : JSON::Any? = nil)
```

PR #334, opened 2022-06-07
create user and query user they must have a return

cacique-coder:
**Description of the change**

Currently search and user creation is returning nil for any case.


w-le (approved):
Yep makes sense to me, not sure why those funcs were defined to return nil

PR #341, opened 2022-06-27
Add Quantum driver

### LINKED ISSUE #320: Driver Request: Lutron Quantum
**Driver Type**

HTTP service

**Manufacturer**

Lutron

**Model/Service**

[Quantum](https://www.lutron.com/asia/Products/Pages/WholeBuildingSystems/Quantum/Overview.aspx)

**Describe any desired functionality**

- Control lights, given address.
- Expose current value of lights, given address

**Additional context**

See: https://drive.google.com/drive/u/1/folders/1NtOtx7PJzu6hedwVyONCs9ESOzJ9tmxt

grkek:
**Description of the change**

Added a Lutron Quantum driver.

**Benefits**

Ability to control Quantum devices.

**Possible drawbacks**

A bug might occur.

**Applicable issues**

- Closes #320

### DIFF
--- a/drivers/lutron/quantum.cr
+++ b/drivers/lutron/quantum.cr
@@ -28,27 +28,32 @@ class Lutron::Quantum < PlaceOS::Driver
   end
 
   def level?(id : Int32)
-    status = @client.try(&.zone.get_status(id))
-    self["area#{id}_level"] = status.try(&.["Level"])
+    status = client.zone.get_status(id)
+    self["area#{id}_level"] = status["Level"]
   end
 
   def level(id : Int32, level : String)
-    @client.try(&.zone.set_status_level(id: id, level: level))
+    client.zone.set_status_level(id: id, level: level)
     self["area#{id}_level"] = level
   end
 
   def scene(id : Int32, scene : Int32)
-    @client.try(&.area.set_scene(id: id, scene: scene))
+    client.area.set_scene(id: id, scene: scene)
     self["area#{id}"] = scene
   end
 
   def scene?(id : Int32)
-    status = @client.try(&.area.get_status(id: id))
-    self["area#{id}"] = status.try(&.["CurrentScene"])
+    status = client.area.get_status(id: id)
+    self["area#{id}"] = status["CurrentScene"]
   end
 
   @[PlaceOS::Driver::Security(Level::Support)]
   def scenes(id : Int32)
-    @client.try(&.area.get_scenes(id: id))
+    client.area.get_scenes(id: id)
+  end
+
+  @[PlaceOS::Driver::Security(Level::Support)]
+  def root
+    client.area.root
   end
 end
--- a/drivers/lutron/quantum.cr
+++ b/drivers/lutron/quantum.cr
@@ -13,7 +13,7 @@ class Lutron::Quantum < PlaceOS::Driver
     device_key: "ab22c585-14e6-4c6b-b418-166728bcc608",
   })
 
-  @client : Client = Client.new(host_name: "", api_key: "", device_key: "")
+  protected getter! client : Client
 
   def on_load
     on_update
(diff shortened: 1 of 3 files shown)


drivers/lutron/quantum.cr
```diff
@@ -0,0 +1,54 @@
...
+  uri_base "https://engineeringwebdemo01.lutron.com/"
+
+  alias Client = ::Quantum::Client
+
+  default_settings({
+    api_key:    "ABCDEFGHIJKLMNOPQRSTUVWXYZ",
+    device_key: "ab22c585-14e6-4c6b-b418-166728bcc608",
+  })
+
+  @client : Client = Client.new(host_name: "", api_key: "", device_key: "")
```
caspiano, line 16: This will raise if accessed before `on_update` is called, but that shouldn't happen. I think it's a bit neater this way.
```suggestion
  protected getter! client : Client
```

grkek: Done.

w-le:
I've tested this and can see it successfully auth with a Quantum sandbox environment. Awesome.

The sandbox environment had no docs on which area ids were available to control.
I see in the Lutron docs that there are additional API endpoints available for discovering what Levels, Areas and Scenes are present.

Are you able to also add those functions to this same driver? Basically the ability to manually query what objects are available for control and return their data, without having to already know what ids are valid.

- List Levels
- List Areas in a given level (or something along those lines)
- List Scenes in a given Area

w-le (dismissed):
looks good thanks

drivers/lutron/quantum.cr
```diff
@@ -0,0 +1,54 @@
...
+    host_name = URI.parse(config.uri.not_nil!).host
+    api_key = setting(String, :api_key)
+    device_key = setting(String, :device_key)
+
+    @client = Client.new(host_name: host_name.not_nil!, api_key: api_key, device_key: device_key)
+  end
+
+  def level?(id : Int32)
+    status = @client.try(&.zone.get_status(id))
+    self["area#{id}_level"] = status.try(&.["Level"])
```
caspiano, line 32:
```suggestion
    status = client.zone.get_status(id)
    self["area#{id}_level"] = status["Level"]
```

drivers/lutron/quantum.cr
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def level(id : Int32, level : String)
+    @client.try(&.zone.set_status_level(id: id, level: level))
```
caspiano, line 36:
```suggestion
    client.zone.set_status_level(id: id, level: level)
```

drivers/lutron/quantum.cr
```diff
@@ -0,0 +1,54 @@
...
+    self["area#{id}_level"] = level
+  end
+
+  def scene(id : Int32, scene : Int32)
+    @client.try(&.area.set_scene(id: id, scene: scene))
```
caspiano, line 41:
```suggestion
    client.area.set_scene(id: id, scene: scene)
```

drivers/lutron/quantum.cr
```diff
@@ -0,0 +1,54 @@
...
+    self["area#{id}"] = scene
+  end
+
+  def scene?(id : Int32)
+    status = @client.try(&.area.get_status(id: id))
+    self["area#{id}"] = status.try(&.["CurrentScene"])
```
caspiano, line 47:
```suggestion
    status = client.area.get_status(id: id)
    self["area#{id}"] = status["CurrentScene"]
```

drivers/lutron/quantum.cr
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  @[PlaceOS::Driver::Security(Level::Support)]
+  def scenes(id : Int32)
+    @client.try(&.area.get_scenes(id: id))
```
caspiano, line 52:
```suggestion
    client.area.get_scenes(id: id)
```

caspiano (changes requested):
I think it would be better to error if the client is not present than to fail silently followed by updating the status variable with potentially invalid data.

grkek:
> I think it would be better to error if the client is not present than to fail silently followed by updating the status variable with potentially invalid data.

It does not compile this way

caspiano:
> It does not compile this way

Using `getter!` should get it to compile

caspiano (approved):

w-le (approved):
lgtm thanks!

w-le:
Hey @grkek are you able to add one additional function to List all available Areas? Used for discovering which area ID corresponds to which physical room name (since this client does not seem to know),  Thanks.

PR #342, opened 2022-07-05
fix(mulesoft): change default ssl settings

chillfox:


drivers/mulesoft/booking_api.cr
```diff
@@ -14,7 +14,7 @@ class MuleSoft::BookingsAPI < PlaceOS::Driver
     time_zone:          "Australia/Sydney",
     ssl_key:            "private key",
     ssl_cert:           "certificate",
-    ssl_auth_enabled:   true,
+    ssl_auth_enabled:   false,
```
caspiano, line 17: What's the reason to disable SSL by default?

chillfox: `on_load` runs with default settings (before settings are set) causing the driver to crash due to invalid ssl_key/ssl_cert

caspiano (approved):

w-le (approved):
ok with me

PR #344, opened 2022-07-06
feat(crestron): fusion driver

chillfox:
Add Crestron Fusion driver.

Based on documentation from 2021 and code samples from 2012.
Models have been left out due to the type of a lot of fields not being documented.
Could use a test system to improve the quality of this driver.

New docs around the Signal Values endpoint are good and it looks like we only need the Signal Values working for now.
Security levels 0 and 1 should be working.


chillfox:
@w-le Do we require security level 2 or is level 1 going to be sufficiently given that we will be using HTTPS?

w-le:
i'll ask the crestron fusion admin

w-le:
They said the "current level" is sufficient, so sounds to me like level 1

stakach (approved):
LGTM

PR #345, opened 2022-07-19
Added Infosilem Campus driver

grkek:
**Description of the change**

Added the Infosilem Campus driver to extract room bookings.

**Benefits**

Driver for the Infosilem Campus.

**Possible drawbacks**

None.

**Additional information**

Additional functions might be required to actually filter bookings by multiple different arguments.

### DIFF
--- a/drivers/infosilem/campus.cr
+++ b/drivers/infosilem/campus.cr
@@ -36,44 +36,24 @@ class Infosilem::Campus < PlaceOS::Driver
     )
   end
 
-  def classes?(room_id : String, start_date : String, end_date : String)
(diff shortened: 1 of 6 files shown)


w-le:
Thanks I tested e2e and it was able to connect to the API ok.
Are you able to add query commands that don't require a param (like building id and room id) that I can use for browsing all available building ids and room ids?

e.g.

list_buildings
list_rooms. (maybe this one can require building id, which I can grab from list_buildings)

w-le:
Happy to merge this into master now @grkek ?

w-le (dismissed):
tested ok e2e

w-le (dismissed):
tested ok e2e

PR #353, opened 2022-08-24
Feat/axiom logic

w-le:
1. adds a super simple room logic that aggregates an array of axiom ids that will be locked/unlocked for the System to which the Logic is deployed to. For use with Trigger Actions. The only reason this logic driver exists is because [Trigger Actions don't utilise System Settings](https://github.com/PlaceOS/triggers/issues/32).

2. Changes the default generic_name of the Axiom driver from "Campus" to "AxiomXa"


grkek (approved):
LGTM

PR #354, opened 2022-08-24
Feat/infosilem logic

w-le:
1. Adds a Logic driver which polls Infosilem Driver for the System's upcoming bookings today, and counts down the minutes until the next event starts. For use with Trigger Conditions to automate actions when an event is about to start.

### DIFF
--- a/drivers/infosilem/campus.cr
+++ b/drivers/infosilem/campus.cr
@@ -48,14 +48,10 @@ class Infosilem::Campus < PlaceOS::Driver
 
     @integration.try(&.call(operation: "EndTransfer", body: end_transfer_body(transfer_id: transfer_id.to_s)))
 
-    
-    self["room_#{room_id}_bookings_#{start_date}_#{end_date}"] = bookings =
-      response
-        .try(&.result
+    response.try(&.result
           .["ObjectData"]
           .["ReservationOccurrences"]
           .["ReservationOccurrence"]?) || [] of Int32
-    bookings
   end
 
   private def start_transfer_options(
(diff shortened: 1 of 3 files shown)


drivers/infosilem/campus.cr
```diff
@@ -48,9 +48,7 @@ class Infosilem::Campus < PlaceOS::Driver
 
     @integration.try(&.call(operation: "EndTransfer", body: end_transfer_body(transfer_id: transfer_id.to_s)))
 
-    self["room_#{room_id}_bookings_#{start_date}_#{end_date}"] =
```
w-le, line 51: I think there might be a core bug where when a Module status variable changes at the same time that an exec response occurs, the exec response is replaced by the status change notification message which is something like "status updated module.status_var = {new value}" which is not valid JSON and so the upstream Logic driver reports an error converting the response to JSON. The issue went away when I removed this status update.

drivers/infosilem/room_schedule.cr
```diff
@@ -0,0 +1,70 @@
...
+    todays_events = Array(Event).from_json(fetch_events(today, today))
+    #todays_events = fetch_events(today, today).map { |e| Event.from_json(e) }
+    @todays_upcoming_events = todays_events.select { |e| e.startTime > Time.local }
+    self[:todays_upcoming_events] = @todays_upcoming_events
+
+    return [] of Event if @todays_upcoming_events.empty?
+
+    next_event = @todays_upcoming_events.min_by { |e| e.startTime }
+    update_event_details(next_event)
+    schedule.clear
```
w-le, line 46: Let me know if you think there's a cleaner way of always having the 15min crons (infosilem requests) while only having the 1mins crons when we know that there are >0 events.

drivers/infosilem/campus.cr
grkek: What can we do about this @stakach ?

grkek (approved):

PR #371, opened 2022-10-06
feat(infosilem/logic): ignore container events

w-le:
An additional request from the users of this RoomSchedule logic driver.

There will be scenarios where events encompass other events. I'll call these "container events". They would like these events to not trigger automation (i.e. be ignored by RoomSchedule).

I think the sort I added here should reduce the processing from n^2 to nlogn.


grkek (approved):
LGTM, keep in mind comparing time values is a bit tricky.

PR #372, opened 2022-10-07
fix(cisco): make more dna_spaces fields optional

chillfox:


grkek (approved):
LGTM, plain nilable values

PR #378, opened 2022-11-03
feat(leviton acquisuite): add webhook driver

camreeves:
Leviton driver for the Acquisuite DAS.

This pretty much just creates a webhook which receives log files and saves them in a status variable to later run influx queries against.

The log files have their fields defined by 'config files' which are also POSTed via a webhook along with a checksum and file time. 

To determine which config files to send the DAS requests a 'manifest' ie. a list of devices and their checksums and file times. This is populated by listening for log file requests and adding any new devices to the manifest. Not sure how often the manifest is requested though which could be an issue (though I assume it's fairly often).


grkek (approved):
LGTM

PR #385, opened 2022-11-10
Swap stock HTTP client with Halite for the Cisco ISE driver

grkek:


stakach:
I don't see what using Halite is adding here?
However using it does mean the driver won't indicate if there are connection issues

stakach:
you can add setting: `https_insecure: true` if you need insecure TLS versions
and setting: `https_verify: "none"` if you don't want to verify tls

grkek:
> I don't see what using Halite is adding here? However using it does mean the driver won't indicate if there are connection issues

Trying to see if swapping the stock client will make a change since I haven't gotten any far with it

PR #387, opened 2022-11-25
feat(cisco/room_kit): default settings enable ppl counting

w-le:
Also the previous default settings would impact video routing which is usually very bad.


w-le:
The same changes have since been made to master separately/directly by others

PR #389, opened 2022-11-30
Update ISE driver with Internal Users

grkek:
These attributes must be pre-defined in the ISE admin panel before we actually start creating internal users since it will raise an error of unknown attributes being passed into the API when creating a user.


stakach:
Still not sure why this isn't implemented as a regular driver?
The library isn't really going to get any use outside of the driver and now we've lost control of the comms and introduced a bunch of new dependencies.

Disadvantages of the library implementation:
1. extra dependencies
2. harder to update (have to update the library and the drivers repo)
3. no proxy support from backoffice
4. can't configure timeouts from backoffice
5. can't configure how to handle TLS layer from backoffice (trusted certs etc)
6. connection status feedback is lost, will always indicate it's connected in backoffice when comms are broken

PR #393, opened 2022-12-12
feat: Add Cisco ISE 3.0 driver

grkek:
Adds the ability to create internal users in the ISE 3.0 API


w-le:
wow names are unique? but many people have the same first and last name - what happens when we need to add a user with a same name as an existing user?

grkek:
> wow names are unique? but many people have the same first and last name - what happens when we need to add a user with a same name as an existing user?

The link below demonstrates how we generate an user name which is unique for each user.

https://github.com/PlaceOS/drivers/blob/07dfcf26d1390f31392f904dddf05577f9b3fed1/drivers/cisco/ise/models/internal_user.cr#L7

An example would be as such:

`internalUser-4eed6c54fed4`

w-le (approved):

PR #403, opened 2022-12-16
feat(lenel/openaccess): reader searching, raw lookup, event pagination

w-le:
Been running this modified version in production fine for a year.

Additional features work fine:
- pagination of events
- card reader searching
- raw lookups

low level alarm related fields of reader were commented out as they will likely never be used. 
high level alarm fields remain as they may be used in the future.

There is an existing behaviour that still needs to be improved. On module start, it won't immediately go into online state. Need to manually force it to check_status before it will go Online.


stakach (approved):
LGTM

PR #405, opened 2022-12-20
feat(placeos/mailers): ISE network creds in Event and Booking emails

w-le:
1. **New Event Mailer logic** driver that **subscribes** to Bookings.bookings in given Zones, filters today's events (other filters can be added in the future) and sends the **Host** (can change target/s in the future) a template email with optional network creds.

2. **Edit existing Booking Notifier logic** driver so that network creds can optionally be generated and included in the Booking notification emails.

3. Edit existing Visitor Mailer logic driver, so that network creds can optionally be generated and included in the emails.
  a. Note that it deliberately only creates/sends the network creds at Guest creation time (event/booking creation time) and not at `send_reminders` cron time, because for security reasons the currently supported network credential provider (Cisco ISE) does not allow us to retrieve the existing user password (so a new random password must be set each time we send the password in an email). It would be very confusing to end users if multiple passwords were sent in multiple emails (creation time, then reminder time) about a single reservation and only 1 was valid. So we deliberately the send the network creds only at creation time.


drivers/place/event_mailer.cr
```diff
@@ -0,0 +1,144 @@
...
+      now = Time.local(location: timezone)
+      event.event_start >= now.at_beginning_of_day && event.event_start <= now.at_end_of_day
+    end
+  end
+
+  private def fetch_network_user(user_email : String, user_name : String)
+    # Check if they already exist
+    response = JSON.parse(network_provider.get_internal_user_by_email(user_email).get.as_s)
+    logger.debug { "Response from Network Identity provider for lookup of #{user_email} was:\n #{response} } " } if @debug
+    return {response["name"], response["password"]} unless response["error"]
```
grkek, line 137: You need to update the password for the returned user since the password is returned as ***** and not clear-text.

grkek (changes requested):

drivers/place/event_mailer.cr
grkek: Use [this](https://github.com/PlaceOS/drivers/blob/master/drivers/cisco/ise/guests.cr#L188) or [this](https://github.com/PlaceOS/drivers/blob/master/drivers/cisco/ise/guests.cr#L202) to update the password of an already existing user.

drivers/cisco/ise/network_access.cr
```diff
@@ -56,64 +59,30 @@ class Cisco::Ise::Guests < PlaceOS::Driver
...
-    # custom_attributes.merge!(@custom_data)
-
-    internal_user.name = username
+    name ||= email
+    password ||= generate_password
+
+    # Not sure if this response is still accurate for ISEv3, todo: check it and align with ISEv3's response
+    # return {"username" => username, "password" => generate_password}.merge(@custom_data) if @test_mode
+
+    # Is there a one line way of doing this?
```
grkek, line 76: You can extend the `{name: name}` NamedTuple and provide the required JSON keys instead of assigning them one by one.

drivers/cisco/ise/network_access.cr
```diff
@@ -199,21 +168,46 @@ class Cisco::Ise::Guests < PlaceOS::Driver
...
+  #   "phoneNumber"        => phone_number,
+  #   "smsServiceProvider" => sms_service_provider.to_s,
+  #   "guestType"          => guest_type,
+  #   "portalId"           => portal_id,
+  # } of String => String
+
+  # custom_attributes.merge!(@custom_data)
+  # end
+
+  # Will be lowercase letters and numbers
```
grkek, line 208: This might be an issue since it needs one uppercase character and a symbol character to be present in the password

drivers/place/booking_notifier.cr
```diff
@@ -365,4 +392,27 @@ class Place::BookingNotifier < PlaceOS::Driver
...
+    {user_email, password}
+  end
+
+  def create_network_user(user_email : String, password : String)
+    response = network_provider.create_internal_user(email: user_email, name: user_email).get
+    logger.debug { "Response from Network Identity provider for creating user #{user_email} was:\n #{response}\n\nDetails:\n#{response.inspect}" } if @debug
+    {response["name"], password}
+  end
+
+  # It's a temporary password that changes each booking, so 6 chars (lowercase and numbers) is fine. We want it to be easy to briefly remember and type
```
grkek, line 413: Add an uppercase character and a symbol here as well

grkek (changes requested):
The code needs to be tested other than the issues I pointed out it LGTM

w-le:
hey @stakach are you free to review this on Tuesday Jan 10?

PR #410, opened 2023-01-12
feat(microsoft/graph): new driver for direct ms graph api calls

w-le:
I chose to use the placeos/office365 lib instead of placeos/calendar (extend the existing graph api calendar driver), as it doesn't appear possible to use `@client.graph_request` when @client is placeos/calendar instead of placeos/office365. 

Is this the right track? Or is there a better approach.

This compiles but is untested.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -28,6 +28,9 @@ dependencies:
   place_calendar:
     github: PlaceOS/calendar
 
+  office365:
+    github: PlaceOS/office365
+
   placeos-compiler:
     github: placeos/compiler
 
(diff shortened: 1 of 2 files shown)


grkek (dismissed):
LGTM

shard.yml
```diff
@@ -28,6 +28,9 @@ dependencies:
   place_calendar:
     github: PlaceOS/calendar
 
+  office365:
```
stakach, line 31: this is included by `place_calendar` so can be removed
pull the latest version using `shards update` instead

stakach (changes requested):

PR #411, opened 2023-01-16
fix(leviton): allow for gzip decompression in log files

camreeves:
Main issue at the moment is that gzip decompression doesn't appear to work inside the specs, but the same logic (from what I can see) works in an isolated environment.

Here's the isolated code that successfully completes the first spec: https://gist.github.com/camreeves/6f66deda7fb891146f9b7118f6505b7a


drivers/leviton/acquisuite_spec.cr
```diff
@@ -2,23 +2,24 @@ require "placeos-driver/spec"
 
 DriverSpecs.mock_driver "Leviton::Acquisuite" do
   headers = {"Content-Type" => ["multipart/form-data; boundary=MIME_BOUNDRY_MIME_BOUNDRY_MIME_BOUNDRY"]}
-  test_devices = ["mb-001", "mb-002"]
+  test_devices = ["mb-001"]
 
   # First, we need to receive some failed LOGFILEUPLOAD requests to work out our list of devices
   test_devices.each do |device_name|
-    dev_log = File.read("/app/repositories/local/drivers/leviton/#{device_name}.log")
+    dev_log = File.read("/app/repositories/local/drivers/leviton/#{device_name}.63BD5AFD_2.log.gz")
```
grkek, line 9: Is this supposed to be specific 63BD5AFD_2?

drivers/leviton/acquisuite_spec.cr
```diff
@@ -2,23 +2,24 @@ require "placeos-driver/spec"
...
     body = create_request(
       "LOGFILEUPLOAD",
       "Temp Inputs / Branch Circuits",
       device_name[-1].to_s,
       "9a6d278642b64db73c754271de733758",
       "2022-09-12 21:25:55",
       "LOGFILE",
-      "modbus/#{device_name}.log",
-      dev_log
+      "modbus/#{device_name}.63BD5AFD_2.log.gz",
```
grkek, line 17: ditto

grkek (changes requested):

drivers/leviton/acquisuite_spec.cr
camreeves: Yeah

Issue #408, opened 2023-01-10
feat(cisco/meraki): support webhook subscriptions for switch port up/down


w-le:
Extend existing Meraki driver to support webhook subscriptions for switch port up and down

Taken from a private internal solution design document authored by Will:
### 3.1.1.2 New Logic Driver: Meraki Switch Port Status
A new PlaceOS Logic Module will initially fetch the state of every switch port that we are interested in (switch port connected to a desk that is being monitored) and expose these as status variables.
The Logic module will use the existing Meraki Dashboard API driver to fetch the data.
Logic Module’s settings (inherited from System/Zones) will include an array of Network Switch serial strings (e.g. "Q2ZP-VE3T-JC2X")
For each switch, get all it’s ports’ status 
[GET /devices/{serial}/switch/ports/statuses](https://developer.cisco.com/meraki/api-latest/#!get-device-switch-ports-statuses)
Example Response
[
    {
        "portId": "1",
        "enabled": true,
        "status": "Disconnected",
        "isUplink": false,
        "errors": [
            "Port disconnected"
        ],
        "warnings": [],
        "speed": "",
        "duplex": "",
        "usageInKb": {
            "total": 0,
            "sent": 0,
            "recv": 0
        },
        "cdp": {
            "platform": "Cisco CISCO1921/K9",
            "deviceId": "CPE-REDACTED-BRISBANE-9272.cpe.overthew",
            "portId": "GigabitEthernet0/1",
            "address": "10.1.74.66",
            "managementAddress": "10.1.74.66",
            "version": "Cisco IOS Software, C1900 Software (C1900-UNIVERSALK9-M), Version 15.6(3)M4, RELEASE SOFTWARE (fc1)\nTechnical Support: http://www.cisco.com/techsupport\nCopyright (c) 1986-2018 by Cisco Systems, Inc.\nCompiled Mon 05-Feb-18 09:56 by prod_rel_team",
            "capabilities": "Router, Source route bridge, Switch"
        },
        "clientCount": 0,
        "powerUsageInWh": 0.0,
        "trafficInKbps": {
            "total": 0.0,
            "sent": 0.0,
            "recv": 0.0
        },
        "securePort": {
            "enabled": false,
            "active": false,
            "authenticationStatus": "Disabled",
            "configOverrides": {}
        }
    },
    {
        "portId": "11",
        "enabled": true,
        "status": "Connected",
        "isUplink": false,
        "errors": [],
        "warnings": [],
[...]

PR #412, opened 2023-01-17
feat(cisco/meraki): add support for wired desk tracking

stakach:
currently missing the requests to obtain the port status with the mac addresses


w-le:
yep that port status request is part of https://github.com/PlaceOS/drivers/issues/408
so best if @grkek adds that as a PR into this branch (`meraki-wired-desks`)

and then at the end we PR meraki-wired-desks into master once.

PR #416, opened 2023-01-20
feat: Add identity groups to creation/update methods of the internal users

grkek:
Updated the creation method from casting a wrong properties and added identity groups


drivers/cisco/ise/network_access.cr
```diff
@@ -186,6 +188,32 @@ class Cisco::Ise::NetworkAccess < PlaceOS::Driver
...
+      "Authorization" => @basic_auth,
+    })
+
+    raise "failed to get internal user by email, code #{response.status_code}\n#{response.body}" unless response.success?
+
+    JSON.parse(response.body)
+  end
+
+  def update_internal_user_identity_groups_by_name(name : String, identity_groups : Array(String))
+    internal_user = get_internal_user_by_name(name)
```
w-le: ise allows internaluser to be updated by name directly, so you can improve this func to be 1 api call instead of 2 (no need to get_internal_user first)
just like `update_internal_user_password_by_name` above

w-le (approved):

PR #417, opened 2023-01-24
feat(visitor_mailer): multiple templates

chillfox:
Add templates for guest events

```cr
event_template:    "event",
booking_template:  "booking",
```


w-le:
nice i that's definitely a step in the right direction though I think one part still remains:
both Resource bookings (like ME booking my desk and receiving a booking confirmation template email) and visitor invites (me inviting an external visitor and THEM getting the visitor template email) are Bookings, yet they need different Templates, and we still need this driver to differentiate between the two (looks like right now they will send the same template).

i can't seem to find any documentation on what a visitor invitation Signal payload is (to see how it differs to a desk booking signal payload)

w-le (approved):

PR #424, opened 2023-02-06
feat(visitor_mailor): all day event time PPT-107

chillfox:
adds a new field for use in templates:

- `event_time`, shows `all day` for all day bookings and `event_start` for other bookings


chillfox:
@stakach should we do this for more types of emails or just these?

stakach (approved):
LGTM

PR #428, opened 2023-02-09
Add GoBright driver PPT-35

grkek:


stakach (changes requested):
The driver should be structured similarly to https://github.com/PlaceOS/drivers/blob/master/drivers/vergesense/vergesense_api.cr

i.e. only expose the relevant API requests and not save anything to state
no need for things like `self["booking"] = client.bookings.create(booking)`
replace with just `client.bookings.create(booking)`

Also the functions should return proper types instead of just JSON::Any
https://github.com/place-technology/bright/blob/master/src/bright/api/bookings.cr#L18-L19

I can see the models have been defined, but are not in use?
It's much simpler to reason about errors and what is being returned if we have types in place.

PR #432, opened 2023-02-14
feat(survey): dedupe emails PPT-111

chillfox:
Only send one email for each unique pair of `email + survey_id` per run, and mark all duplicates as sent.

### DIFF
--- a/drivers/place/survey_mailer_spec.cr
+++ b/drivers/place/survey_mailer_spec.cr
@@ -1,0 +1,83 @@
+require "placeos-driver/spec"
+require "placeos-driver/interface/mailer"
+
+class StaffAPI < DriverSpecs::MockDriver
+  def get_survey_invites(survey_id : Int64? = nil, sent : Bool? = nil)
+    survey_id ||= 1
+
+    unsent_invites = [
+      {
+        id:        1,
+        survey_id: survey_id,
+        token:     "QWERTY",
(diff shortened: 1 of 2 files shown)


chillfox:
It would be good to get https://github.com/PlaceOS/drivers/pull/430 merged as well (adds tests around survey invites to the staff API driver)

drivers/place/survey_mailer_spec.cr
```diff
@@ -0,0 +1,83 @@
...
+    to : String | Array(String),
+    template : Tuple(String, String),
+    args : TemplateItems,
+    resource_attachments : Array(ResourceAttachment) = [] of ResourceAttachment,
+    attachments : Array(Attachment) = [] of Attachment,
+    cc : String | Array(String) = [] of String,
+    bcc : String | Array(String) = [] of String,
+    from : String | Array(String) | Nil = nil
+  )
+    self[:sent] = self[:sent].as_i + 1
```
grkek, line 57: This simply counts the amount sent?

grkek (approved):

drivers/place/survey_mailer_spec.cr
chillfox: yeah,
I can make it store the email if you think that's better

grkek: I think count is quite enough for now

PR #433, opened 2023-02-16
fix(docker-compose): update redis and install-shards image to multi-arch images tag

naqvis:
**Description of the change**

Update `docker-compose` file to use images which support multi-architecture.

**Benefits**

Developers running on AMD or ARM64 machines should be able to use images specific for their architecture.

**Possible drawbacks**

  None


stakach (approved):
LGTM

PR #435, opened 2023-03-06
fix(leviton): ensure config list is loaded from settings

camreeves:
Pretty simple fix. I just forgot to load the config list out of settings when the driver starts/updates.


stakach (approved):
LGTM

PR #440, opened 2023-03-14
Add Vecos driver PPT-73

grkek:


stakach (changes requested):
The driver should be structured similarly to https://github.com/PlaceOS/drivers/blob/master/drivers/vergesense/vergesense_api.cr

i.e. only expose the relevant API requests and not save anything to state
no need for things like `self["lock_bank_status"] = client.locker_banks.get_status(bank_id)`
replace with just `client.locker_banks.get_status(bank_id)`

Also the functions should return proper types instead of just JSON::Any
We have to build a [location services driver](https://github.com/PlaceOS/drivers/blob/master/drivers/vergesense/location_service.cr) on top of this so need to know what the responses look like and be able to detect errors via the compiler.
Right now to develop a location services driver we would have to look at the documentation for each request to work out what is being returned.

PR #441, opened 2023-03-24
Add 25Live Pro driver PPT-106

grkek:


stakach (changes requested):
Can we convert this to something similar to the [gobright](https://github.com/PlaceOS/drivers/tree/master/drivers/gobright) driver

api.cr
api_spec.cr
models/* (a folder with the models in it)

grkek:
> Can we convert this to something similar to the [gobright](https://github.com/PlaceOS/drivers/tree/master/drivers/gobright) driver
> 
> api.cr api_spec.cr models/* (a folder with the models in it)

Sure we can, I will convert it ASAP

stakach:
Why ParameterBuilder? Seem like it's reinventing the wheel a bit when something like `URI::Params` does the same thing
https://github.com/PlaceOS/drivers/blob/master/drivers/vecos/releezme.cr#L232-L243

grkek:
> > Why ParameterBuilder? Seem like it's reinventing the wheel a bit when something like `URI::Params` does the same thing https://github.com/PlaceOS/drivers/blob/master/drivers/vecos/releezme.cr#L232-L243
> 
> I think that in this case reinventing the wheel saves you much more time than you actually have to spend on differentiating each variable if it is an Array, Hash or a basic type.
> 
> From my experience I can say that this way it is much easier to work with Query/Form parameters, URI::Params is too primitive for me.
> 
> ### Parameter Builder
> ```crystal
> io = IO::Memory.new
> builder = ParameterBuilder.new(io)
> 
> builder.add("pagingTake", "100")
> builder.add("LocationId", location) if location
> builder.add("spaceTypes", types)
> ```
> 
> #### VS
> ### URI::Params
> ```crystal
> params = URI::Params.build do |form|
>   form.add "pagingTake", "100"
>   form.add "LocationId", location.to_s unless location.presence.nil?
>   if types
>     types = types.is_a?(Array) ? types : [types]
>     types.each do |type|
>       form.add "SpaceTypes", type.value.to_s
>     end
>   end
> end
> ```
> 
> I get the point you are explaining to me, however I also see a clear code readability and automation possibility of not having to deal with certain types of objects. Alas it would be great if we had the ParameterBuilder embedded into utils or as such.

After taking a look at the source code of the STD version, it seems like it is doing pretty much the same.

stakach (approved):
LGTM - thanks!

stakach:
worth running `crystal tool format` before merging though

PR #445, opened 2023-05-08
fix(leviton): add SUCCESS to body of responses

camreeves:
Simply added the word "SUCCESS" or "FAILURE" to each body except the manifest request.


stakach (approved):
LGTM

PR #449, opened 2023-06-19
feat: Update the Metasys driver with additional endpoints PPT-614

grkek:


MarioFavalliR (commented):
All looks good to me in terms of endpoints and what we need. The last thing which is more a "nice to have" rather than a "must" is to change `equipment_ids` variable's name (line 17 and 30) to `id` in the Driver settings. This is because we are not using only equipment ids but also object and network devices ids. But this is only a nice to have so NOT URGENT

chillfox (approved):
looks good

PR #455, opened 2023-07-11
Fix/johnson control/send command headers

MarioFavalliR:
**Description of the change**

Added content_type to headers for put request when send command is executed
 per api documentation
**Benefits**

The api was failing before this quick fix

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**


grkek (dismissed):
LGTM

PR #457, opened 2023-08-03
fix(releezme/api): auth domain is a setting not hardcoded

w-le:
The auth api changes depending on dev/stg/prod environments, so needs to be a setting not a hardcoded string


stakach (approved):
LGTM

PR #459, opened 2023-09-06
feat(globalcache): Add projector screen logic driver

w-le:
enable meeting room control logic driver to control projector screens via globalcache (not just IP modules


w-le:
tested OK in prod months ago

PR #461, opened 2023-09-21
feat(placeos/room-events): add list_users func

w-le:
Some role based accounts running the Booking panel app (via api key) may not have MS Graph API delegated permissions to query the AAD user directory (e.g. PlaceOS local accounts).

This change allows those apps to query users via a driver exec instead of via a staff-api call. A corresponding frontend change would also be required.

### DIFF
--- a/drivers/place/bookings.cr
+++ b/drivers/place/bookings.cr
@@ -40,10 +40,10 @@ class Place::Bookings < PlaceOS::Driver
     room_image: "https://domain.com/room_image.svg",
     sensor_mac: "device-mac",
 
-    hide_meeting_details: false,
-    hide_meeting_title:   false,
+    hide_meeting_details:      false,
+    hide_meeting_title:        false,
     enable_end_meeting_button: false,
-    max_user_search_results: 20,
+    max_user_search_results:   20,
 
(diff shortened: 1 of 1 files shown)


drivers/place/bookings.cr
```diff
@@ -208,6 +211,11 @@ class Place::Bookings < PlaceOS::Driver
     schedule.in(1.seconds) { poll_events } unless (subscription = @subscription) && !subscription.expired?
   end
 
+  # Allow apps to search for attendees (to add to new bookings) via driver instead of via staff-api (as some role based accounts may not have MS Graph access)
+  def list_users(query : String? = nil, limit : UInt32? = 20_u32)
+    calendar.list_users(query, limit).get.as_a
```
stakach, line 216: you don't actually need to call `.get.as_a` unless the function is used within the driver
this will be done automatically with more performance for any remote requests

```suggestion
    calendar.list_users(query, limit)
```

stakach (changes requested):

drivers/place/bookings.cr
w-le: Thanks! didn't know that

w-le:
tested e2e ok

PR #462, opened 2023-09-21
fix(placeos/visitor-mailer): determine host name with calendar driver…

w-le:
This change allows kiosk apps using API Keys associated with placeos local role based accounts (no MS Graph api delegated access) to determine the host name, using an AAD lookup via calendar driver instead of staff-api driver (delegated permissions).

Note that before deploying this visitor mailer driver update to any clients that REQUIRE the hostname lookup to use staff-api instead of calendar driver, be sure to set setting:
`determine_host_name_using: "staff-api-driver"`
to ensure that it functions as it used to.


stakach (approved):
LGTM

PR #465, opened 2023-10-13
feat(gobright/api): add GET spaces/state request

w-le:
Add support for GET spaces/state: 
https://t1b.gobright.cloud/swagger/index.html?url=/swagger/v1/swagger.json#/SpacesV2/SpacesV2_GetState

Mostly copied/pasted from existing GET spaces request (above it). Changed q param names only.

ALSO

Allow Gobright location services driver to specify a default spacetype for device_locations(). Just to workaround a new bug in GoBright API where no data is returned (instead of all) when no space type filter is specified.


grkek (dismissed):
LGTM

PR #481, opened 2023-10-31
fix(delta): PROJ-428 Convert ValueProperty class fields from mandatory to optional

grkek:


w-le (approved):
thx, tested e2e working ok with temperature,  humidity abnd PM2.5 properties

w-le:
oh wait but we don't get the actual value of the property (which is the purpose of that function)
can you also add to our model, optional fields that match the real response we saw?
like
```
"present-value":{"$base":"Real","value":"22.54"},"status-flags":{"$base":"BitString","value":""},"event-state":{"$base":"Enumerated","value":"normal"},"out-of-service":{"$base":"Boolean","value":0},"units":{"$base":"Enumerated","value":"\u00b0C"},"description":{"$base":"String","value":"Temperature"},"reliability":{"$base":"Enumerated","value":"no-fault-detected"}}
```

PR #495, opened 2024-05-11
feat(PPT-1303): Update Smartalock with the locker interface

grkek:


drivers/floorsense/lockers_websocket.cr
```diff
@@ -0,0 +1,977 @@
...
+  # return `nil` or `{"location": "wireless", "assigned_to": "bob123", "mac_address": "abcd"}`
+  def check_ownership_of(mac_address : String) : OwnershipMAC?
+    logger.debug { "smartalock is incapable of checking ownerships" }
+    nil
+  end
+
+  def device_locations(zone_id : String, location : String? = nil)
+    logger.debug { "smartalock is incapable to locate devices" }
+    [] of Nil
+  end
```
stakach, line 940: do we need this code as it no longer implements the locatable interface

drivers/floorsense/lockers_websocket.cr
```diff
@@ -0,0 +1,977 @@
...
+    else
+      raise "bad response result (#{resp.code}) #{resp.message}"
+    end
+  end
+
+  # ========================================
+  # Lockers Interface
+  # ========================================
+
+  class PlaceLocker
```
stakach, line 752: this is defined in the locatable interface:
https://github.com/PlaceOS/driver/blob/master/src/placeos-driver/interface/lockers.cr#L9

you should use `alias PlaceLocker = PlaceOS::Driver::Interface::Lockers::PlaceLocker`
and can monkey patch it if you want to add some new fields

Similar to what is done in [releezme](https://github.com/PlaceOS/drivers/blob/d82c7bab9fa9fc57d50a816590f075897a19e65a/drivers/vecos/releezme_locations.cr#L10)

grkek: Currently doing that, if you scroll up you can see the alias.

drivers/floorsense/lockers_websocket.cr
grkek: Seems like it was required by the actual interface.

stakach: ahh yes! Could you please implement the `device_locations` code similar to what is happening on releezme

Needs to output data like this:
https://docs.placeos.com/how-to/location-services/location-services#lockers

grkek: I don't think it supports such a case, I will investigate.

stakach: effectively this function needs to grab all the currently active locker bookings (like bookings that are live right now, excluding past and future bookings) and filter by level or building zones

so the zone_ids need to be converted - basically very similar to [this code, just for lockers](https://github.com/PlaceOS/drivers/blob/master/drivers/floorsense/location_service.cr#L154-L221).

So you'll need to [maintain mappings, like this code](https://github.com/PlaceOS/drivers/blob/master/drivers/floorsense/location_service.cr#L19-L37)

I still think you should have followed the recommendation from the PPT to keep the locker location service separate from the drivers code (basically just create a `locker_location_service.cr`) that's the same as `location_service.cr` just for lockers
image

as this would have allowed the settings in `location_service.cr` to be shared with `locker_location_service.cr` - only one place where the zone_ids => floorsense id mappings need to be maintained

grkek: I did see that comment but it was a bit late, thank you, will do!

stakach (approved):
LGTM

PR #503, opened 2024-06-11
fix(auto_release): handle prefered time crossing midnight in utc, correct day_of_week numbering, handle last minute changes to bookings

chillfox:


grkek (approved):
LGTM

PR #505, opened 2024-06-12
fix(auto_release): use #enabled? for all enabled checks

chillfox:


drivers/place/auto_release.cr
```diff
@@ -144,7 +144,11 @@ class Place::AutoRelease < PlaceOS::Driver
       if preferences = get_user_preferences?(booking.user_id)
         day_of_week = Time.unix(booking.booking_start).day_of_week.value
         day_of_week = 0 if day_of_week == 7 # Crystal uses 7 for Sunday, but we use 0 (all other days match up)
+
+        # convert unix timestamp to float hours/minutes
+        # e.g. 7:30AM = 7.5
         event_time = Time.unix(booking.booking_start).hour + (Time.unix(booking.booking_start).minute / 60.0)
```
stakach, line 150: we should probably only do this a single time and re-use the result
```crystal
booking_start = Time.unix(booking.booking_start)
event_time = booking_start .hour + (booking_start.minute / 60.0)
```

stakach (changes requested):
Looks good after minor change

drivers/place/auto_release.cr
chillfox: done

stakach (approved):
LGTM

PR #512, opened 2024-08-27
feat(template_mailer): send email using metadata templates

chillfox:


drivers/place/template_mailer.cr
```diff
@@ -0,0 +1,224 @@
...
+    on_update
+  end
+
+  def on_update
+    @cache_timeout = setting?(Int64, :cache_timeout) || 300_i64
+
+    org_zone_ids.each do |zone_id|
+      update_template_fields(zone_id)
+    end
+  end
```
chillfox, line 43: @stakach 

Does it make sense to have it update the available fields on every org zone it can find, or should it be limited to where it's located `(zone_ids & system.zones).first`?

How about scheduling an update to avoid having to restart it every time a new driver with `template_fields` is added/removed

drivers/place/auto_release.cr
```diff
@@ -221,6 +224,24 @@ class Place::AutoRelease < PlaceOS::Driver
     self[:released_booking_ids] = [] of Int64
   end
 
+  def template_fields : Array(TemplateFields)
+    # TODO: add more fields to be consistent with other templates
```
chillfox, line 228: More useful fields needs to be added to this driver as it's missing most of the fields available to other drivers.

drivers/place/visitor_mailer.cr
```diff
@@ -312,6 +315,60 @@ class Place::VisitorMailer < PlaceOS::Driver
...
+      {name: "event_title", description: "Title or purpose of the visit"},
+      {name: "event_start", description: "Start time (e.g., #{time_now.to_s(@time_format)})"},
+      {name: "event_date", description: "Date of the visit (e.g., #{time_now.to_s(@date_format)})"},
+      {name: "event_time", description: "Time of the visit (or 'all day' for 24-hour events)"},
+    ]
+
+    invitation_fields = common_fields + [
+      {name: "room_name", description: "Name of the room or area being visited"},
+      {name: "network_username", description: "Network access username (if network credentials enabled)"},
+      {name: "network_password", description: "Generated network access password (if network credentials enabled)"},
```
chillfox, line 335: We could make this more dynamic and only include the user/pass if `send_network_credentials` is enabled.

drivers/place/template_mailer.cr
chillfox: I added a cron schedule for updating

drivers/place/template_mailer.cr
```diff
@@ -0,0 +1,243 @@
...
+
+  descriptive_name "Template Mailer"
+  generic_name :Mailer
+  description %(uses metadata templates to send emails via the SMTP mailer)
+
+  default_settings({
+    cache_timeout: 300, # timeout for the template cache
+    timezone:    "GMT",
+    update_schedule:    "*/20 * * * *", # cron schedule for updating template fields
+  })
```
chillfox, line 21: Would it make sense to add an option to only update fields, so not delete fields from drivers that are no longer available.
Or make them sticky for x number of updates before being removed, to avoid any intermittent issues.

chillfox: added an option to make the fields sticky for x number of updates

drivers/place/template_mailer.cr
chillfox: made it only update the org its in

stakach (approved):
LGTM

PR #521, opened 2024-10-17
fix(PPT-1540): Update settings with proper names

grkek:


stakach (commented):
this is a breaking change though... we'd need to investigate the impact on clients who have the old setting configured

grkek:
> this is a breaking change though... we'd need to investigate the impact on clients who have the old setting configured

I can put in a clause which will handle both of these

drivers/place/bookings.cr
```diff
@@ -32,6 +32,7 @@ class Place::Bookings < PlaceOS::Driver
 
     application_permissions:    true,
     include_cancelled_bookings: false,
+    show_qr_code:               false,
     hide_qr_code:               false,
```
stakach, line 36: We should remove the old setting from the defaults
Otherwise looks good

stakach (changes requested):

drivers/place/bookings.cr
```diff
@@ -164,7 +164,11 @@ class Place::Bookings < PlaceOS::Driver
 
     self[:custom_qr_color] = setting?(String, :custom_qr_color)
     self[:custom_qr_url] = setting?(String, :custom_qr_url)
-    self[:show_qr_code] = !(setting?(Bool, :hide_qr_code) || false)
+
+    hide_qr_code = setting?(Bool, :hide_qr_code) || false
+    show_qr_code = setting?(Bool, :show_qr_code) || false
+
+    self[:show_qr_code] = !hide_qr_code || show_qr_code
```
stakach, line 171: The settings both kind of clash
i.e. by default hide_qr_code == true, but if show_qr_code is set to false then the setting won't be applied.

```suggestion
    self[:show_qr_code] = show_qr_code.nil? ? !hide_qr_code : show_qr_code
```

drivers/place/bookings.cr
```diff
@@ -164,7 +164,11 @@ class Place::Bookings < PlaceOS::Driver
 
     self[:custom_qr_color] = setting?(String, :custom_qr_color)
     self[:custom_qr_url] = setting?(String, :custom_qr_url)
-    self[:show_qr_code] = !(setting?(Bool, :hide_qr_code) || false)
+
+    hide_qr_code = setting?(Bool, :hide_qr_code) || false
+    show_qr_code = setting?(Bool, :show_qr_code) || false
```
stakach, line 169:
```suggestion
    show_qr_code = setting?(Bool, :show_qr_code)
```

stakach (changes requested):
needs to be backwards compatible and the settings can't clash

PR #522, opened 2024-11-07
Fix/harness

viv-4:
**Description of the change**

Updates usage of `docker-compose` to `docker compose`

**Benefits**

`docker compose` is available in `docker` since version 20.10
it is no longer common to install `docker-compose` separately
some github ci containers require this change

**Possible drawbacks**

If a user of harness has a docker version < 20.10 they will need to update docker


stakach (approved):
LGTM

PR #531, opened 2024-12-19
feat(visitor_mailer): [PPT-1430] notify on induction accepted/declined

chillfox:
requres https://github.com/PlaceOS/staff-api/pull/337 for the signals to actually  be sent


stakach (approved):
LGTM

PR #532, opened 2025-01-07
feat(auto_release_locker): [PPT-1725] auto release lockers on specific interval

naqvis:


drivers/place/auto_release_locker.cr
```diff
@@ -0,0 +1,79 @@
...
+    [] of Booking
+  end
+
+  def release_lockers
+    bookings = get_bookings
+    released = 0
+    bookings.each do |booking|
+      logger.debug { "releasing booking #{booking.id} as it is within the time_after window" }
+      begin
+        staff_api.update_booking(booking_id: booking.id, booking_end: Time.utc.to_unix, checked_in: false)
```
stakach, line 71: you need to add `.get` at the end of this line if you are interested in the result of the operation (and want to log the error)
Otherwise it's request and forget

stakach: Also as you haven't included the `def update_booking` function in your mocks, this would error with the current spec if you do add `.get`
Probably worth doing both of these

stakach (changes requested):
Nice! Only one minor change

stakach (approved):
LGTM

PR #538, opened 2025-02-05
feat(auto_release): [PPT-1835] skip_created_after_start and skip_same_day settings

chillfox:
add config options:
- skip_created_after_start: true
- skip_same_day: false


stakach (changes requested):
I actually think we should automate the timezone config using:
```
def on_update
  @timezone = nil
  # ...
end

  protected getter timezone : Time::Location do
    tz = config.control_system.try(&.timezone) || "UTC"
    Time::Location.load(tz)
  end
```
that way we don't need to configure it in settings

stakach (approved):
LGTM

PR #539, opened 2025-02-05
feat(bookings): [PPT-1823] use event visibility to control exposure

naqvis:


stakach (changes requested):
Looks good, let's not use shards override though

stakach (approved):
LGTM

PR #540, opened 2025-02-12
feat(room_at_capacity_mailer): [PPT-1721|PPT-1852] notify when people_count equals or exceeds capacity

chillfox:


stakach (changes requested):
This will send an email every 20 seconds while the room is at capacity
We probably want to do 2 things:

1. track where rooms are at or over capacity (system_id => over_capactity_detected_count)
    increment the count every poll
2. only send an email when the over_capactity_detected_count == 2 (i.e. 5min polling means email sent when over capacity for 10min)
3. if a room drops below capacity then delete the system_id from the hash or set to 0

We might also want to ensure only one email is sent per meeting?
but i think the above should work well enough

chillfox:
There is a `debounce_time_minutes` option that limits how frequently notifications can go out for the same system.
requiring it to be over for multiple checks is a good idea, I will add that now.

drivers/place/room_at_capacity_mailer.cr
```diff
@@ -0,0 +1,148 @@
...
+    on_update
+  end
+
+  @notify_email : Array(String) = [] of String
+  @email_template : String = "room_at_capacity"
+  @debounce_time_minutes : Int32 = 60
+  @over_capactity_detected_count : Int32 = 2
+
+  @last_email_sent : Hash(String, Time) = {} of String => Time
+  @over_capacity : Hash(String, Int32) = {} of String => Int32
```
stakach, line 46:
```suggestion
  @over_capacity : Hash(String, Int32) = Hash(String, Int32).new { |hash, sys_id| hash[sys_id] = 0 }
```

drivers/place/room_at_capacity_mailer.cr
```diff
@@ -0,0 +1,148 @@
...
+      system_ids.each do |system_id|
+        sys = system(system_id)
+        next unless sys.exists?("Bookings", 1)
+        next unless sys.capacity > 0
+
+        if people_count = sys.get("Bookings", 1).status?(Int32, "people_count")
+          logger.debug { "people count for #{system_id}: #{people_count}" }
+
+          if people_count >= sys.capacity
+            @over_capacity[system_id] = @over_capacity[system_id]? ? @over_capacity[system_id] + 1 : 1
```
stakach, line 73:
```suggestion
            @over_capacity[system_id] += 1
```

stakach (changes requested):
minor change

PR #542, opened 2025-02-17
feat(cloud_xapi): PPT-1818 added webx cloud xapi driver

naqvis:


w-le:
Thanks @naqvis , looks great!
One change we will need though is that the `device_id` will not be a static string in this Drivers Settings (such that an instance of this driver will only over control 1 device), but a String Parameter of all commands this Driver exposes, such that the caller (e.g. a Logic driver or frontend app) can request this Service Driver to command any of hundreds of physical devices that the xAPI session allows.
Are you able to make that change?

w-le (changes requested):
Thanks @naqvis , looks great!
One change we will need though is that the device_id will not be a static string in this Drivers Settings (such that an instance of this driver will only over control 1 device), but a String Parameter of all commands this Driver exposes, such that the caller (e.g. a Logic driver or frontend app) can request this Service Driver to command any of hundreds of physical devices that the xAPI session allows.
Are you able to make that change?

w-le (approved):
device-id change looks good, thanks. PR approved and I'll test e2e then comment or merge.

In the mean time, you can continue to work on Phase 2 of PPT-1818, thanks

PR #543, opened 2025-03-03
feat(cloud_xapi): PPT-1818 phase#2 added UI extension commands

naqvis:


w-le (approved):
Tested working ok. Got 400 for a known invalid device id on exec led_colour?

PR #550, opened 2025-03-28
fix(floorsense/desk_list): handle controllers lacking desk api

viv-4:
https://acaprojects.atlassian.net/browse/SD-5653


stakach (approved):
LGTM

PR #553, opened 2025-03-31
Add logic driver to sync Cisco Webex Nagivator LED color to Bookings status via xAPI. Untested

w-le:
todo: 
- DONE: compile without errors, 
- spec

### DIFF
--- a/drivers/cisco/booking_panel_led_sync.cr
+++ b/drivers/cisco/booking_panel_led_sync.cr
@@ -1,0 +1,49 @@
+require "placeos-driver"
+
+class Cisco::BookingPanelLedSync < PlaceOS::Driver
+  descriptive_name "Cisco Webex Navigator Panel LED Sync"
+  generic_name :Navigator_LED_Sync
+  description "Sync Cisco Webex Navigator Panel LED to Bookings.in_use status"
+
+  default_settings({
+    led_color_when_room_booked: "Red",
+    led_color_when_room_available: "Green",
+    webex_panel_device_id: "Ensure this is set in each System's settings"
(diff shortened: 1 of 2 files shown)


drivers/cisco/booking_panel_led_sync.cr
```diff
@@ -0,0 +1,49 @@
...
+
+    system.subscribe("Bookings_1", "in_use") do |_sub, value|
+      next unless ["true", "false"].includes?(value)
+      self[:room_in_use] = room_in_use = value == "true"
+      set_led_color(room_in_use)
+    end
+  end
+
+  def sync_led_color_now
+    return unless ["true", "false"].includes?(room_in_use = system[:Bookings][:in_use].as_bool)
```
w-le, line 40: Driver Harness states that system[:Bookings][:in_use] is JSON::Any and errors with and withohut `.as_bool`
What's the best way to handle this to fetch the Bool value of another module's status?

chillfox: What's the actual error you are getting?

You can try inserting these lines above it to get more details about what `system[:Bookings][:in_use]` actually contains.
```
pp! system[:Bookings][:in_use]
pp! system[:Bookings][:in_use] == nil
pp! typeof(system[:Bookings][:in_use])
```

w-le:
tested e2e working ok against live xAPI driver and Bookings module for the 2 main cases Bookings.in_use true/false

w-le:
merged for deployment. But I will keep this feat branch open for reviews of the specs that I will write later on.

PR #554, opened 2025-03-31
feat(floorsense/desks_websocket): split desk & locker sync

viv-4:
Split locker and desk sync methods
Add settings for sync interval

https://acaprojects.atlassian.net/browse/SD-5656


stakach (commented):
I think we can close this as resolved by this change
https://github.com/PlaceOS/drivers/commit/d524553593068be4e3acd365f0c5c6d0b3a68c4d

the merged function ensures fewer API requests are made

PR #556, opened 2025-05-05
feat(place/meet): power on preview_outputs on startup

w-le:
When the system is started, power on all devices listed in the `preview_outputs` setting.
This needs to be tested when joined and unjoined, before rolling out.


w-le:
awaiting UAT

w-le:
tested and approved by client

PR #560, opened 2025-05-30
feat: PPT-2065 Sennheiser TCCM driver

naqvis:


stakach (changes requested):
Looks great!
It should implement the mutable interface too
https://github.com/PlaceOS/driver/blob/master/src/placeos-driver/interface/muteable.cr

drivers/sennheiser/team_connect_ceiling_medium.cr
```diff
@@ -0,0 +1,99 @@
...
+  include Interface::AudioMuteable
+  include Sennheiser::Models
+  descriptive_name "TeamConnect Ceiling Medium"
+  generic_name :TCCM
+
+  uri_base "https://device_ip"
+
+  default_settings({
+    username:      "api",
+    password:      "",
```
stakach, line 15:
```suggestion
    basic_auth: {
      username: "api",
      password: "",
    },
```

you don't actually need to handle auth in the headers, the driver will do it for you

stakach (changes requested):
minor change to how auth is handled

drivers/sennheiser/team_connect_ceiling_medium.cr
```diff
@@ -0,0 +1,99 @@
...
+    device_ip:     "192.168.0.1",
+    debug_payload: false,
+  })
+
+  @debug_payload : Bool = false
+  @username : String = ""
+  @password : String = ""
+
```
stakach, line 23:
```suggestion
def on_load
  transport.before_request do |request|
    logger.debug { "performing request: #{request.method} #{request.path}" }
  end

  on_update
end
```

stakach:
added the SSCVv1 protocol driver
and a second crack at the SSCVv2 driver with subscription support

PR #562, opened 2025-06-11
feat(sony): add Bravia REST API driver

KesterJJ:
Add comprehensive Sony Bravia display driver supporting REST API control with:
- Power management (on/off/standby with status polling)
- Volume control (set, mute/unmute, volume up/down with state polling)
- Input switching (HDMI1-4, Component, Composite, etc. with current input polling)
- Extended functionality (apps, picture settings, IR codes, system info)
- Thread-safe API calls with mutex protection
- Robust error handling and PSK authentication validation
- Full interface compliance (Powerable, Muteable, Switchable)
- Comprehensive test suite with edge case coverage


w-le:
Very cool, does it compile and pass the spec? Using driver test harness

sony/displays/bravia_rest_spec.cr
```diff
@@ -0,0 +1,419 @@
+require "placeos-driver/spec"
+
+DriverSpecs.mock_driver "Sony::Displays::BraviaRest" do
+  settings({
+    psk: "test1234",
+  })
+
+  # Power Control Tests
+  it "should power on the display" do
```
stakach, line 9: These `it` methods won't work here, but I'll update the driver spec runner so it works as expected

stakach (changes requested):

stakach:
I updated it by fixing the spec runner on the AI repo
then asking to finish testing this branch

stakach (dismissed):
LGTM

PR #565, opened 2025-06-20
Feat/new driver/sony/ptz cgi

KesterJJ:


w-le:
Using placeos/drivers test harness, currently the ptz_cgi.cr driver compiles, but fails the spec run with:
```
{
    "error": "In drivers/sony/camera/ptz_cgi_spec.cr:3:13

 3 | DriverSpecs.mock_driver \"Sony::Camera::PtzCGI\" do
                 ^----------
Error: instantiating 'DriverSpecs.mock_driver(String)'


In drivers/sony/camera/ptz_cgi_spec.cr:3:13

 3 | DriverSpecs.mock_driver \"Sony::Camera::PtzCGI\" do
                 ^----------
Error: instantiating 'DriverSpecs.mock_driver(String)'


In drivers/sony/camera/ptz_cgi_spec.cr:18:3

 18 | update_settings({\"uri_base\" => \"http://192.168.1.100\"})
      ^--------------
Error: undefined method 'update_settings' for DriverSpecs (with ... yield) and top-level (current scope)
"
}
```

However I have just noticed that there is actually already a 4 year cgi_protocol.cr driver in the same folder, by Steve. No idea if this has ever been used, but we have 2 options now. I will e2e test with the older cgi_protocol.cr driver first.

PR #574, opened 2025-08-13
feat: [PPT-2181] Added Solution2 PMS Driver

naqvis:


w-le (approved):
perfect. service and logic driver tested e2e working ok

PR #582, opened 2025-09-03
Updated ZR-CSAPI and NVX RX

kvu-ucla:
- Updated ZR-CSAPI to add method for stopping wireless sharing
- adding updated switch local with added USB-C1 and USB-C2 inputs


w-le (approved):
Looks good and safe, thank you!

PR #603, opened 2026-04-09
feat(public_events): [PPT-2247] filter public events

chillfox:
For use with https://github.com/PlaceOS/rest-api/pull/430


stakach (approved):
LGTM

PR #607, opened 2026-04-30
feat(visitor_mailer): [PPT-2375] notify after edit

chillfox:
Depends upon this staff-api change: https://github.com/PlaceOS/staff-api/pull/370


stakach (approved):
LGTM

PR #614, opened 2026-06-16
fix(visitor_mailer): reduce duplicate emails (PPT-2375)

chillfox:


naqvis (commented):
LGTM.
Except that harness script changes. I doubt this hack will work during cases where redis get restarted. Also don't believe problem is related to musl binaries but the base image or other docker configurations issues.

chillfox:
> LGTM. Except that harness script changes. I doubt this hack will work during cases where redis get restarted. Also don't believe problem is related to musl binaries but the base image or other docker configurations issues.

I have removed the harness changes

naqvis (approved):
LGTM

PR #621, opened 2026-07-30
fix(visitor_mailer): follow the host reassigned on an event (PPT-2375)

chillfox:
Scenario 1 of PPT-2375: *"the 'Booking Host Changed' notification should be sent to the original host. Instead the new host is receiving the Visitor Invite Confirmation email."* — the mailer half.

The reassignment itself is PlaceOS/staff-api#383, which makes `host` on every signal mean *who is hosting* and reports the mailbox owner separately as `organiser_email`. With that merged the mailer's existing logic already notifies the previous host and filters the new one; this parses the new field, pins the behaviour with a test, and deals with the other half of QA's report.

## The new host receiving a visitor invitation

They were not being treated as a host at all — they were a newly added attendee, and the signal still named the old host, so `skip_host_email` had nothing to match. staff-api#383 fixes that case, but it is the visible symptom of something broader: front ends mark **every** attendee of a meeting as an expected visitor, so colleagues are announced exactly like external guests and receive visitor invitations and QR codes.

`skip_internal_domain_email` (default `false`) drops anyone sharing the host's email domain, on invitations, check-in and induction notifications, and change notifications. It is the no-configuration counterpart to `host_domain_filter`, which needs an explicit domain list.

Left off by default: staff visiting *another* site are legitimate visitors and share the host's domain, so turning this on everywhere would silently stop those emails. Worth enabling per system — including the dev server, where it is what stops QA seeing the invitation to the new host.

Emails aimed at the host (`notify_checkin`, `notify_original_host`) are unaffected — the filter tests the attendee's domain, not the recipient's.

## Testing

4 specs: `host_domain_filter` and `skip_internal_domain_email` suppressing a colleague's invitation and their change notification while an external visitor still gets both, the setting being off by default, and a reassignment signal notifying the previous host while naming the new host on the visitor's change email.

`./harness report --basic-render --verbose --tests=1 drivers/place/visitor_mailer_spec.cr` passes; `crystal tool format` and `ameba` clean.

Stacked on #620. Happy to split `skip_internal_domain_email` out into its own change if you would rather keep this to the signal shape alone — it is independent of the rest.

### DIFF
--- a/drivers/place/visitor_mailer.cr
+++ b/drivers/place/visitor_mailer.cr
@@ -71,6 +71,12 @@ class Place::VisitorMailer < PlaceOS::Driver
     # unaffected.
     skip_host_email: true,
 
+    # When true, attendees whose email domain matches the host's are treated as
+    # colleagues rather than visitors and are not emailed. Front ends tend to
+    # mark every attendee as an expected visitor, so staff invited to a meeting
+    # would otherwise receive visitor invites and QR codes.
+    skip_internal_domain_email: false,
+
     domain_uri:      "https://example.com/",
     jwt_private_key: PlaceOS::Model::JWTBase.private_key,
   })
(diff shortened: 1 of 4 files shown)


stakach (approved):
LGTM

PR #629, opened 2026-08-27
fix(visitor_mailer): (PPT-2375)

chillfox:
Don't merge without checking with Sharmila that it does not introduce new bugs.


camreeves:
Closing this one as superseded. Everything here went to master through #631 (squash 6e62d22530) together with the follow-up fixes from the HIO UAT testing, and the squash credits you as co-author, so nothing is lost. Thanks for the groundwork on the building-zone timezone handling.

PR #630, opened 2026-09-22
chore(staff_api.cr): use ingress nginx service as default URI

viv-4:
Uses the usual cluster internal ingress nginx Service as the default URI
Adds a placeholder setting for the host header
Clarifies unused value in initialization of place_domain


stakach (approved):
LGTM

